PR #10, opened 2022-08-29
Local Bucketing Implementation

JamieSinn:
Done:
- Sets up local bucketing tests for each of the new functions
- Events callback system for readying and notifications
- Config manager

WIP:
- Cleanup of user exposed properties/functions.

Still needed:
- Event polling impl assuming WASM handles it.
- More tests


localbucketing.go
```diff
@@ -62,33 +90,204 @@ func (d *DevCycleLocalBucketing) Initialize() (err error) {
 	}
 	d.wasmMemory = d.wasmInstance.GetExport(d.wasmStore, "memory").Memory()
 
-	return nil
+	log.Println("Initialized DevCycle LocalBucketing")
```
jonathannorris, line 93: cleanup logs?

localbucketing.go
```diff
@@ -62,33 +90,204 @@ func (d *DevCycleLocalBucketing) Initialize() (err error) {
...
+	defer d.wasmMutex.Unlock()
+	tokenAddr, err := d.newAssemblyScriptString(d.sdkKey)
+	if err != nil {
+		return
+	}
+	optionsAddr, err := d.newAssemblyScriptString(options)
+	if err != nil {
+		return
+	}
+	_generateBucketedConfigForUser := d.wasmInstance.GetExport(d.wasmStore, "initEventQueue").Func()
```
jonathannorris, line 130: rename `_generateBucketedConfigForUser` as `_initEventQueue`

localbucketing.go
```diff
@@ -62,33 +90,204 @@ func (d *DevCycleLocalBucketing) Initialize() (err error) {
...
+	d.flushMutex.Unlock()
+}
+func (d *DevCycleLocalBucketing) flushEventQueue() (payload []FlushPayload, err error) {
+	d.wasmMutex.Lock()
+	defer d.wasmMutex.Unlock()
+	tokenAddr, err := d.newAssemblyScriptString(d.sdkKey)
+	if err != nil {
+		return
+	}
+	_generateBucketedConfigForUser := d.wasmInstance.GetExport(d.wasmStore, "flushEventQueue").Func()
```
jonathannorris, line 149: rename `_generateBucketedConfigForUser`

localbucketing.go
```diff
@@ -62,33 +90,204 @@ func (d *DevCycleLocalBucketing) Initialize() (err error) {
...
+}
+
+func (d *DevCycleLocalBucketing) checkEventQueueSize() (length int, err error) {
+	d.wasmMutex.Lock()
+	defer d.wasmMutex.Unlock()
+	tokenAddr, err := d.newAssemblyScriptString(d.sdkKey)
+	if err != nil {
+		return
+	}
+	_generateBucketedConfigForUser := d.wasmInstance.GetExport(d.wasmStore, "eventQueueSize").Func()
```
jonathannorris, line 166: rename `_generateBucketedConfigForUser`

jonathannorris (changes requested):

api_devcycle.go
```diff
@@ -154,19 +185,45 @@ func (a *DVCClientService) AllFeatures(ctx context.Context, body UserData) (map[
...
 @return Variable
 */
-func (a *DVCClientService) Variable(ctx context.Context, body UserData, key string, defaultValue interface{}) (Variable, error) {
+func (a *DVCClientService) Variable(ctx context.Context, userdata UserData, key string, defaultValue interface{}) (Variable, error) {
+	if !a.client.DevCycleOptions.DisableLocalBucketing {
+		bucketed, err := a.generateBucketedConfig(userdata)
+
+		variableEvaluationType := ""
+		if bucketed.Variables[key].IsDefaulted {
+			variableEvaluationType = EventType_VariableEvaluated
```
jonathannorris, line 200: Where are these defined? can we make sure they are `aggVariableEvaluated` and `aggVariableDefaulted`

api_devcycle.go
```diff
@@ -154,19 +185,45 @@ func (a *DVCClientService) AllFeatures(ctx context.Context, body UserData) (map[
...
+		} else {
+			variableEvaluationType = EventType_VariableDefaulted
+		}
+		if !a.client.DevCycleOptions.DisableAutomaticEventLogging {
+			err = a.queueAggregateEvent(bucketed, DVCEvent{
+				Type_:  variableEvaluationType,
+				Target: key,
+			})
+			if err != nil {
+				return Variable{}, err
```
jonathannorris, line 210: We should not return any errors here, we do not want to block a variable for an event-saving issue.

api_devcycle.go
```diff
@@ -154,19 +185,45 @@ func (a *DVCClientService) AllFeatures(ctx context.Context, body UserData) (map[
...
+			}
+		}
+		if err != nil {
+			return Variable{}, err
+		}
+		return bucketed.Variables[key], err
```
jonathannorris, line 216: Is there a way to ignore all these errors? basically if there is an error we should return a Variable with the default value as a backup, this method shouldn't return any errors.

api_devcycle.go
```diff
@@ -296,11 +353,16 @@ func (a *DVCClientService) Variable(ctx context.Context, body UserData, key stri
...
- * @param body
+  - @param ctx context.Context - for authentication, logging, cancellation, deadlines, tracing, etc. Passed from http.Request or context.Background().
+  - @param body
+
 @return map[string]Variable
 */
 func (a *DVCClientService) AllVariables(ctx context.Context, body UserData) (map[string]Variable, error) {
+	if !a.client.DevCycleOptions.DisableLocalBucketing {
+		user, err := a.generateBucketedConfig(body)
+		return user.Variables, err
```
jonathannorris, line 364: Same here this probably shouldn't return errors.

api_devcycle.go
```diff
@@ -435,19 +497,33 @@ func (a *DVCClientService) AllVariables(ctx context.Context, body UserData) (map
...
+
 @return InlineResponse201
 */
-func (a *DVCClientService) Track(ctx context.Context, user UserData, event Event) (InlineResponse201, error) {
+
+func (a *DVCClientService) Track(ctx context.Context, user UserData, event DVCEvent) (InlineResponse201, error) {
+	if a.client.DevCycleOptions.DisableCustomEventLogging {
+		return InlineResponse201{}, nil
+	}
+	event.ClientDate = time.Now()
```
jonathannorris, line 510: I don't think this needs to be set here, AS will handle this.

api_devcycle.go
```diff
@@ -435,19 +497,33 @@ func (a *DVCClientService) AllVariables(ctx context.Context, body UserData) (map
...
+	event.Type_ = "customEvent"
```
jonathannorris, line 511: I don't think this is correct either. Basically the only param that should be need to be set on the DVCEvent JSON Obj that is sent to AS is `type`: 
image

The rest of the types are optional: 
image

jonathannorris (changes requested):

example/cloud/main.go
```diff
@@ -0,0 +1,53 @@
...
+		Key: environmentKey,
+	})
+	dvcOptions := devcycle.DVCOptions{
+		EnableEdgeDB:                 false,
+		DisableLocalBucketing:        true,
+		EventsFlushInterval:          time.Second * 10,
+		PollingInterval:              time.Second * 10,
+		RequestTimeout:               time.Second * 10,
+		DisableAutomaticEventLogging: false,
+		DisableCustomEventLogging:    false,
```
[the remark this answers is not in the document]
JamieSinn: These are used to configure the eventqueueoptions object - but I'll just pass them in and skip checking them in the queue methods.

eventqueue.go
```diff
@@ -0,0 +1,139 @@
...
+		if err != nil {
+			return err
+		}
+		err = e.localBucketing.initEventQueue(string(str))
+		ticker := time.NewTicker(e.options.EventsFlushInterval)
+
+		go func(ctx context.Context) {
+			for {
+				select {
+				case <-ctx.Done():
```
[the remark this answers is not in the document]
JamieSinn: application crash, or context is failed. This prevents the app from hanging because there's a background thread running. It's also called to abort polling if there is a non-retryable error.

localbucketing.go
JamieSinn: C# Has this too - mostly a "things worked" info message
Can remove if not needed.

eventqueue.go
```diff
@@ -0,0 +1,139 @@
...
+	return nil
+}
+
+func (e *EventQueue) QueueEvent(user UserData, event DVCEvent) error {
+	if q, err := e.checkEventQueueSize(); err != nil || q {
+		fmt.Println(err)
+		log.Println("event queue is full. Dropping event")
+		return fmt.Errorf("event queue is full. Dropping event")
+	}
+	if !e.options.DisableLocalBucketing {
```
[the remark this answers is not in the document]
JamieSinn: These are checks to prevent null pointers if accessing an unintialized localbucketing object.

eventqueue.go
```diff
@@ -0,0 +1,139 @@
...
+	"log"
+	"net/http"
+	"time"
+)
+
+func (e *EventQueue) initialize(options *DVCOptions, localBucketing *DevCycleLocalBucketing) error {
+	e.context = context.Background()
+	e.httpClient = http.DefaultClient
+	e.options = options
+	if !e.options.DisableLocalBucketing && localBucketing != nil {
```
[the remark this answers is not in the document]
JamieSinn: These are checks to prevent null pointers if accessing an unintialized localbucketing object.

api_devcycle.go
JamieSinn: Did a hybrid - returning a default and the error to not break procedure.

api_devcycle.go
JamieSinn: Logged the error and ignored it.

api_devcycle.go
JamieSinn: Returned a default if there was an error - which is just an empty string:variable map. Returning the error is needed still, but not catastrophic since the user can still ignore it.

api_devcycle.go
JamieSinn: removed.

api_devcycle.go
JamieSinn: Fixed.

configuration.go
```diff
@@ -49,11 +50,36 @@ type APIKey struct {
...
+	PollingInterval              time.Duration `json:"pollingInterval,omitempty"`
+	RequestTimeout               time.Duration `json:"requestTimeout,omitempty"`
+	DisableAutomaticEventLogging bool          `json:"disableAutomaticEventLogging,omitempty"`
+	DisableCustomEventLogging    bool          `json:"disableCustomEventLogging,omitempty"`
+	MaxEventQueueSize            int           `json:"maxEventsPerFlush,omitempty"`
+	FlushEventQueueSize          int           `json:"minEventsPerFlush,omitempty"`
 }
 
-type Configuration struct {
+func (o *DVCOptions) CheckDefaults() {
```
jonathannorris, line 64: Check the logic on these because some of them we try and set minumum values like this: 
image

jonathannorris (changes requested):

eventqueue.go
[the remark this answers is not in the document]
JamieSinn: yeah can invert.

api_devcycle.go
```diff
@@ -154,20 +177,47 @@ func (a *DVCClientService) AllFeatures(ctx context.Context, body UserData) (map[
...
- * @param body
- * @param key Variable key
+  - @param ctx context.Context - for authentication, logging, cancellation, deadlines, tracing, etc. Passed from http.Request or context.Background().
+  - @param body
+  - @param key Variable key
+
 @return Variable
 */
-func (a *DVCClientService) Variable(ctx context.Context, body UserData, key string, defaultValue interface{}) (Variable, error) {
+func (a *DVCClientService) Variable(ctx context.Context, userdata UserData, key string, defaultValue interface{}) (Variable, error) {
```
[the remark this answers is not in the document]
JamieSinn: We should always be including an error, regardless of whether the variable is a valid return. It's just a golang handler convention.

api_devcycle.go
```diff
@@ -435,19 +487,33 @@ func (a *DVCClientService) AllVariables(ctx context.Context, body UserData) (map
...
+
 @return InlineResponse201
 */
-func (a *DVCClientService) Track(ctx context.Context, user UserData, event Event) (InlineResponse201, error) {
+
+func (a *DVCClientService) Track(ctx context.Context, user UserData, event DVCEvent) (InlineResponse201, error) {
+	if a.client.DevCycleOptions.DisableCustomEventLogging {
+		return InlineResponse201{}, nil
+	}
+	if event.Type_ == "" {
```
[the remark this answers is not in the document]
JamieSinn: Type is a reserved keyword, and if started with `_Type` - it would not be an "exported" value from the struct - ie - not accessible outside of the package.

api_devcycle.go
```diff
@@ -435,19 +487,33 @@ func (a *DVCClientService) AllVariables(ctx context.Context, body UserData) (map
...
- * @param ctx context.Context - for authentication, logging, cancellation, deadlines, tracing, etc. Passed from http.Request or context.Background().
- * @param body
+  - @param ctx context.Context - for authentication, logging, cancellation, deadlines, tracing, etc. Passed from http.Request or context.Background().
+  - @param body
+
 @return InlineResponse201
 */
-func (a *DVCClientService) Track(ctx context.Context, user UserData, event Event) (InlineResponse201, error) {
+
+func (a *DVCClientService) Track(ctx context.Context, user UserData, event DVCEvent) (InlineResponse201, error) {
```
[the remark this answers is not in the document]
JamieSinn: It's a generated return type - realistically we can abstract it to a bool.

configmanager.go
```diff
@@ -1,15 +1,121 @@
...
+	log.Printf("Config set. ETag: %s\n", e.configETag)
+	if e.firstLoad {
+		e.firstLoad = false
+		log.Println("DevCycle SDK Initialized.")
+	}
+	return nil
+}
 
+func (e *EnvironmentConfigManager) getConfigURL() string {
+	return fmt.Sprintf("https://config-cdn.devcycle.com/config/v1/server/%s.json", e.environmentKey)
```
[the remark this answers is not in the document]
JamieSinn: We can - it's planned but wasn't part of this initial runthrough

model_platformdata.go
```diff
@@ -0,0 +1,30 @@
+package devcycle
+
+type PlatformData struct {
```
jonathannorris, line 3: can you add hostname to PlatformData, and have it automatically set from the systems hostname.

jonathannorris: ie: 
image

jonathannorris (approved):

PR #12, opened 2023-01-12
chore: upgrade wasmtime lib

nsalamad:


go.mod
```diff
@@ -3,7 +3,7 @@ module github.com/devcyclehq/go-server-sdk
 go 1.19
 
 require (
-	github.com/bytecodealliance/wasmtime-go v0.39.0
+	github.com/bytecodealliance/wasmtime-go/v4 v4.0.0
```
JamieSinn, line 6: can we actually pin this to the same version as C#?

nsalamad: yup

PR #16, opened 2023-01-17
fix: use httpconfiguration base URLs for events and config manager

ajwootto:
Fix the default value for the events queue by passing around the HttpConfiguration object derived from the client options, which contains the correct values for the base CDN and events API URLs.

### DIFF
--- a/configmanager.go
+++ b/configmanager.go
@@ -19,17 +19,15 @@ type EnvironmentConfigManager struct {
 	context        context.Context
 	cancel         context.CancelFunc
 	httpClient     *http.Client
-	httpConfig     *HTTPConfiguration
-	options        *DVCOptions
 }
 
-func (e *EnvironmentConfigManager) Initialize(environmentKey string, options *DVCOptions) (err error) {
-	e.options = options
+func (e *EnvironmentConfigManager) Initialize(environmentKey string, localBucketing *DevCycleLocalBucketing) (err error) {
+	e.localBucketing = localBucketing
 	e.environmentKey = environmentKey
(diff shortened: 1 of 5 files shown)


configmanager.go
```diff
@@ -119,10 +117,8 @@ func (e *EnvironmentConfigManager) setConfig(response *http.Response) error {
 }
 
 func (e *EnvironmentConfigManager) getConfigURL() string {
-	configBasePath := "https://config-cdn.devcycle.com"
-	if e.options.ConfigCDNOverride != "" {
-		configBasePath = e.options.ConfigCDNOverride
-	}
+	configBasePath := e.localBucketing.cfg.ConfigCDNBasePath
+
 	return fmt.Sprintf("%s/config/v1/server/%s.json", configBasePath, e.environmentKey)
```
JamieSinn, line 122: We should allow overriding the full uri too

ajwootto: why?

JamieSinn: https://github.com/DevCycleHQ/dotnet-server-sdk/blob/main/DevCycle.SDK.Server.Local/ConfigManager/EnvironmentConfigManager.cs#L104

ajwootto: if we want that capability for the SDKs we have to add it to all of them and agree on the standard name for the option. What's the usecase for it?

JamieSinn: Where you have a proxy that you're contacting instead of the actual CDN path.

JamieSinn (approved):

PR #22, opened 2023-01-17
Move InitalizeLocalBucketing call into NewDVCClient

emmawillis:


client.go
```diff
@@ -63,13 +63,6 @@ type service struct {
 }
 
 func InitializeLocalBucketing(environmentKey string, options *DVCOptions) (ret *DevCycleLocalBucketing, err error) {
```
JamieSinn, line 65: Because we're no longer exposing this to public users; it should be renamed with a lowercase starting letter to prevent access publicly.

client.go
```diff
@@ -84,7 +77,7 @@ func InitializeLocalBucketing(environmentKey string, options *DVCOptions) (ret *
 
 // NewDVCClient creates a new API client.
 // optionally pass a custom http.Client to allow for advanced features such as caching.
-func NewDVCClient(environmentKey string, options *DVCOptions, localBucketing *DevCycleLocalBucketing) (*DVCClient, error) {
+func NewDVCClient(environmentKey string, options *DVCOptions) (*DVCClient, error) {
 	if options.EnableCloudBucketing && environmentKey == "" {
```
ajwootto, line 81: you gotta remove the EnableCloudBucketing check here for the key validation

emmawillis: oh true, oops - fixed

client.go
emmawillis: done!

client.go
```diff
@@ -84,11 +77,11 @@ func InitializeLocalBucketing(environmentKey string, options *DVCOptions) (ret *
...
 // NewDVCClient creates a new API client.
 // optionally pass a custom http.Client to allow for advanced features such as caching.
-func NewDVCClient(environmentKey string, options *DVCOptions, localBucketing *DevCycleLocalBucketing) (*DVCClient, error) {
-	if options.EnableCloudBucketing && environmentKey == "" {
+func NewDVCClient(environmentKey string, options *DVCOptions) (*DVCClient, error) {
+	if environmentKey == "" {
 		return nil, fmt.Errorf("Missing environment key! Call NewDVCClient with a valid environment key.")
 	}
-	if options.EnableCloudBucketing && !environmentKeyIsValid(environmentKey) {
+	if !environmentKeyIsValid(environmentKey) {
```
JamieSinn, line 84: this could be collapsed into the statement above

ajwootto: the error message is different

JamieSinn: Are they not fundamentally the same? Invalid vs missing are both the same issue.

ajwootto: we distinguish them in the other SDKs we've done so far. Invalid means you passed something but it wasn't a real token, missing means you passed nothing

JamieSinn (approved):

PR #25, opened 2023-01-17
chore: added ReadOnlyVariable and made number default values into float64

jsalaber:


model_variable.go
```diff
@@ -9,6 +9,19 @@
...
 
 type Variable struct {
+	ReadOnlyVariable
+	// Default variable value can be a string, number, boolean, or JSON
+	DefaultValue interface{} `json:"defaultValue"`
+	// Identifies if variable was returned with the default value
+	IsDefaulted bool `json:"isDefaulted,omitempty"`
+}
+
+type FeatureVariation struct {
```
ajwootto, line 19: how come this type is defined in the `model_variable` file? is there a more appropriate place for it?

update-bucketing-lib.sh
```diff
@@ -1,5 +1,5 @@
 #!/bin/bash
-BUCKETING_LIB_VERSION="1.0.18"
+BUCKETING_LIB_VERSION="1.1.25"
```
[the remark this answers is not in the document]
jsalaber: @JamieSinn made this change, but I believe it was to use the current version that was released, no changes were made otherwise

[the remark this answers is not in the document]
ajwootto: oh nvm I missed the wasm file change in the diff

model_variable.go
```diff
@@ -9,6 +9,19 @@
 package devcycle
 
 type Variable struct {
+	ReadOnlyVariable
```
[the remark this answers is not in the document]
jsalaber: yeah, it allows the usage of calling `ReadOnlyVariable` properties on `Variable`

model_variable.go
jsalaber: moved it

api_devcycle.go
```diff
@@ -106,15 +107,23 @@ DVCClientService Get variable by key for user data
 @return Variable
 */
 func (a *DVCClientService) Variable(ctx context.Context, userdata DVCUser, key string, defaultValue interface{}) (Variable, error) {
-	defaultRetVal := Variable{Value: defaultValue, Key: key, IsDefaulted: true}
+	convertedDefaultValue := convertDefaultValueType(defaultValue)
+	readOnlyVariable := ReadOnlyVariable{Key: key, Value: convertedDefaultValue}
+	variable := Variable{ReadOnlyVariable: readOnlyVariable, DefaultValue: convertDefaultValueType, IsDefaulted: true}
```
ajwootto, line 112: so does this magically assign all the readOnlyVariable properties then?

when I first read this I thought this was assigning a field called "ReadOnlyVariable" on the variable object, which contained a "ReadOnlyVariable" object

jsalaber: yeah unfortunately inheritance doesn't work as nicely, you can't assign specifically the properties inherited from `ReadOnlyVariable` on the `Variable` object, you have to pass in an object of `ReadOnlyVariable`

ajwootto: thats fine as long as the end-user interface of `Variable` is still like `Variable.Value` and not something like `Variable.ReadOnlyVariable.Value`

model_variable.go
ajwootto, line 19: this needs to be deleted now

api_devcycle.go
```diff
@@ -106,15 +107,23 @@ DVCClientService Get variable by key for user data
...
+		sameTypeAsDefault := compareTypes(bucketed.Variables[key].Value, convertedDefaultValue)
 		variableEvaluationType := ""
-		if bucketed.Variables[key].IsDefaulted {
+		if bucketed.Variables[key].Value != nil && sameTypeAsDefault {
+			variable.Value = bucketed.Variables[key].Value
+			variable.IsDefaulted = false
 			variableEvaluationType = EventType_AggVariableEvaluated
 		} else {
+			if !sameTypeAsDefault {
+				log.Println("Type of Variable does not match DevCycle configuration. Using default value")
```
ajwootto, line 125: any way to have this match the log on the other SDKs, or is it not possible to get a string representation of the types?
image

ajwootto: yeah I think on dotnet we didnt have immediate access to the type names because the check was done elsewhere

api_devcycle.go
```diff
@@ -360,3 +366,36 @@ func (a *DVCClientService) handleError(r *http.Response, body []byte) (err error
...
+	case uint:
+		return float64(value.(uint))
+	case uint8:
+		return float64(value.(uint8))
+	case uint16:
+		return float64(value.(uint16))
+	case uint32:
+		return float64(value.(uint32))
+	case uint64:
+		return float64(value.(uint64))
```
JamieSinn, line 395: You may have an issue where typecasting between unsigned and signed datatypes are platform specific - and can introduce unexpected bugs.

jsalaber: that's what the test harness is for!

JamieSinn: The test harness won't fix this issue - it's specific to the OS/Arch combo.

jsalaber: do you have a concrete example in mind? i'm of the opinion that bugs caused by the language itself due to differing platforms are not our main goal to fix

JamieSinn: https://medium.com/a-journey-with-go/go-cast-vs-conversion-by-example-26e0ef3003f0

This is absolutely something we need to be handling for, otherwise we are failing to return the correct datatype.

jsalaber: that article doesn't mention anything about platform specific quirks for conversion/typecasting

In any case, we can write in our docs that we implicitly convert to float64 for consistency across all number values

JamieSinn: I would put a ticket in to properly convert eventually; maybe a guild ticket

jsalaber: alrighty, can you encapsulate those details in a ticket?

PR #26, opened 2023-01-18
add oninitialized channel that allows the user to make calls before the client is initialized

elliotCamblor:
How it works:
- The user can choose to pass in an "OnInitializedChannel" into DVCOptions
- If this channel exists then the initialization process will be run in a go routine, meaning client calls can be made before the bucketing lib is initialized
- The user can listen on the channel to block the main process until init is complete
- By not passing in the channel param the sdk will behave the same as it did before

### DIFF
--- a/api_devcycle.go
+++ b/api_devcycle.go
@@ -9,8 +9,8 @@ import (
 	"log"
 	"net/http"
 	"net/url"
-	"strings"
 	"reflect"
+	"strings"
 )
 
 var (
@@ -52,8 +52,14 @@ DVCClientService Get all features by key for user data
 func (a *DVCClientService) AllFeatures(ctx context.Context, body DVCUser) (map[string]Feature, error) {
 
 	if !a.client.DevCycleOptions.EnableCloudBucketing {
-		user, err := a.generateBucketedConfig(body)
-		return user.Features, err
+		if a.client.isInitialized {
+			user, err := a.generateBucketedConfig(body)
+			return user.Features, err
+		} else {
+			log.Println("AllFeatures called before client initialized")
+			return map[string]Feature{}, nil
+		}
+
 	}
 	var (
 		httpMethod          = strings.ToUpper("Post")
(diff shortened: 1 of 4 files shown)


JamieSinn (changes requested):
No unit tests?

elliotCamblor:
the whole point of the test harness is so we dont have to write unit tests

JamieSinn (approved):

api_devcycle.go
```diff
@@ -109,28 +114,32 @@ func (a *DVCClientService) Variable(ctx context.Context, userdata DVCUser, key s
 	defaultRetVal := Variable{Value: defaultValue, Key: key, IsDefaulted: true}
 
 	if !a.client.DevCycleOptions.EnableCloudBucketing {
-		bucketed, err := a.generateBucketedConfig(userdata)
+		if a.client.isInitialized {
```
ajwootto, line 117: it would be easier to read if you just early returned here instead

elliotCamblor: word

client.go
```diff
@@ -75,6 +76,26 @@ func initializeLocalBucketing(environmentKey string, options *DVCOptions) (ret *
...
+		if options.OnInitializedChannel != nil {
+			options.OnInitializedChannel <- true
+		}
+		return nil, err
+	}
+	c.localBucketing = localBucketing
+	c.configManager = c.localBucketing.configManager
+	c.eventQueue = c.localBucketing.eventQueue
+	c.isInitialized = true
+	if options.OnInitializedChannel != nil {
```
ajwootto, line 92: does this need to wait until the config is fetched or is that happening implicitly somewhere?

example/local/main.go
```diff
@@ -33,6 +34,10 @@ func main() {
 		log.Printf("Key:%s, feature:%s", key, feature)
 	}
 
+	if <-onInitialized {
```
ajwootto, line 37: do you actually care if the value here is "true", otherwise I don't think you need the if statement you can just wait on the channel directly:
`<-onInitialized`
`log.Printf("client initialized")`

api_devcycle.go
```diff
@@ -234,10 +249,13 @@ func (a *DVCClientService) Track(ctx context.Context, user DVCUser, event DVCEve
 	}
 
 	if !a.client.DevCycleOptions.EnableCloudBucketing {
-		err := a.client.eventQueue.QueueEvent(user, event)
-		return err == nil, err
+		if a.client.isInitialized {
+			err := a.client.eventQueue.QueueEvent(user, event)
+			return err == nil, err
+		} else {
+			return false, errors.New("client not yet initialized")
```
ajwootto, line 256: technically we just log and return on other platforms rather than throwing

ajwootto (approved):

PR #37, opened 2023-01-19
remove auth param shenanigans from sdk

elliotCamblor:


jsalaber (approved):

api_devcycle.go
```diff
@@ -271,7 +258,7 @@ func (a *DVCClientService) Track(user DVCUser, event DVCEvent) (bool, error) {
 	// body params
 	postBody = &body
 
-	r, rBody, err := a.performRequest(a.client.auth, path, httpMethod, postBody, headers, queryParams)
+	r, rBody, err := a.performRequest(nil, path, httpMethod, postBody, headers, queryParams)
```
ajwootto, line 261: seems sus that you have to pass nil here, can't you just remove this requirement from that method too?

elliotCamblor: ill take a look. I was a bit afraid to fuck with that method because its purely autogen'd. Likely a feasible and good change though

elliotCamblor: yea actually I think it just has a bunch of different auth methods we dont care about on it

client.go
```diff
@@ -289,34 +283,6 @@ func (c *DVCClient) prepareRequest(
...
-		}
-
-		// Basic HTTP Authentication
-		if auth, ok := ctx.Value(ContextBasicAuth).(BasicAuth); ok {
-			localVarRequest.SetBasicAuth(auth.UserName, auth.Password)
-		}
-
-		// AccessToken Authentication
-		if auth, ok := ctx.Value(ContextAccessToken).(string); ok {
-			localVarRequest.Header.Add("Authorization", "Bearer "+auth)
```
ajwootto, line 316: this looks important

elliotCamblor: I thought so too, which is why I was initially hesitant to remove the ctx param altogether

elliotCamblor: but `nil` is being passed in literally every place this method is getting called

ajwootto: is the auth header being added somewhere else then?

PR #42, opened 2023-01-20
Only set err field on 400-level cloud errors

emmawillis:


api_devcycle.go
```diff
@@ -353,12 +353,19 @@ func (a *DVCClientService) handleError(r *http.Response, body []byte) (err error
...
+	if len(body) > 0 {
+		err = decode(&v, body, r.Header.Get("Content-Type"))
+		if err != nil {
+			newErr.error = err.Error()
+			return newErr
+		}
 	}
 	newErr.model = v
+
+	if r.StatusCode >= 500 {
```
jsalaber, line 365: should this not go before decoding?

emmawillis: this way the body is decoded when we log it so it should log a clearer error message (i think)

jsalaber: how does it get here if we return the `newErr` on line 360 above?

emmawillis: It should only return newErr if there was an error with actually decoding the body, otherwise err is nil and the body pointer just gets updated so it skips the return statement

jsalaber (approved):

PR #43, opened 2023-01-20
Add default values for platform and sdkType

emmawillis:


api_devcycle.go
```diff
@@ -239,6 +239,8 @@ func (a *DVCClientService) Track(user DVCUser, event DVCEvent) (bool, error) {
 		return false, errors.New("event type is required")
 	}
 
+	user.fillDefaults()
```
ajwootto, line 242: this needs to happen anywhere that user data is used, incl. for evaluations of variables

jsalaber (approved):

model_user_data.go
```diff
@@ -49,3 +49,13 @@ type UserFeatureData struct {
...
 	FeatureVars map[string]string
 }
+
+func (user *DVCUser) fillDefaults() {
+	if user.Platform == "" {
+		user.Platform = "Go"
+	}
+
+	if user.SdkType == "" {
+		user.SdkType = "server"
```
ajwootto, line 59: this should also include sdkVersion and the other fields that we are populating on the other sdks

api_devcycle.go
ajwootto, line 242: we should probably also try and differentiate the user-facing `DVCUser` from the internal one that has the extra autopopulated fields, similar to how we have `DVCUser` and `DVCPopulatedUser` on other platforms.

ajwootto (approved):

jsalaber (approved):

PR #45, opened 2023-01-20
fix: compare type for differing default values

jsalaber:


client.go
```diff
@@ -246,14 +246,16 @@ func (c *DVCClient) Variable(userdata DVCUser, key string, defaultValue interfac
...
-		return localVarReturnValue, err
+		return variable, err
 	}
 
 	if r.StatusCode < 300 {
 		// If we succeed, return the data, otherwise pass on to decode error.
 		err = decode(&localVarReturnValue, body, r.Header.Get("Content-Type"))
-		if err == nil {
-			return localVarReturnValue, err
+		if err == nil && localVarReturnValue.Value != nil && compareTypes(localVarReturnValue.Value, convertedDefaultValue) {
```
ajwootto, line 255: in the case where the compareTypes check is false, this will fall through to the error response handling below. You may not want that, since the request "succeeded". Instead you should probably just check this inside the `err==nil` conditional block and then either return the default value, or not.

client.go
```diff
@@ -246,14 +250,25 @@ func (c *DVCClient) Variable(userdata DVCUser, key string, defaultValue interfac
...
 	if r.StatusCode < 300 {
 		// If we succeed, return the data, otherwise pass on to decode error.
 		err = decode(&localVarReturnValue, body, r.Header.Get("Content-Type"))
-		if err == nil {
-			return localVarReturnValue, err
+		if err == nil && localVarReturnValue.Value != nil {
+			if compareTypes(localVarReturnValue.Value, convertedDefaultValue) {
+				variable.Value = localVarReturnValue.Value
+				variable.IsDefaulted = false
+			} else if localVarReturnValue.Value != nil {
```
ajwootto, line 263: this has to be true here, this condition is redundant

ajwootto (approved):

PR #46, opened 2023-01-20
fix FlushEvents not handling error codes correctly

nsalamad:


nsalamad:
* main
  * **PR #46** <a href="https://app.graphite.dev/github/pr/DevCycleHQ/go-server-sdk/46" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/DevCycleHQ/go-server-sdk/46?utm_source=stack-comment).

eventqueue.go
```diff
@@ -137,26 +137,29 @@ func (e *EventQueue) FlushEvents() (err error) {
...
-					log.Println(err)
-					continue
-				}
-			}
 			log.Println(err)
 			continue
 		}
 
+		if resp.StatusCode >= 500 && resp.StatusCode < 600 {
+			err = e.localBucketing.onPayloadFailure(payload.PayloadId, true)
```
ajwootto, line 154: how come we dont want to mark a payload as failed when there's a request err?

eventqueue.go
```diff
@@ -137,26 +137,29 @@ func (e *EventQueue) FlushEvents() (err error) {
...
+			if err != nil {
```
ajwootto, line 155: is this case possible?

eventqueue.go
```diff
@@ -137,26 +137,29 @@ func (e *EventQueue) FlushEvents() (err error) {
...
-				if err != nil {
-					log.Println(err)
-					continue
-				}
-			}
 			log.Println(err)
 			continue
 		}
 
+		if resp.StatusCode >= 500 && resp.StatusCode < 600 {
```
ajwootto, line 153: I dont think you need to check for <600 since status codes only go up 5xx

eventqueue.go
```diff
@@ -137,26 +137,29 @@ func (e *EventQueue) FlushEvents() (err error) {
 	for _, payload := range payloads {
 		var req *http.Request
 		var resp *http.Response
-		requestBody, err := json.Marshal(BatchEventsBody{Batch: payload.Records})
-		req, err = http.NewRequest("POST", eventsHost+"/v1/events/batch", bytes.NewReader(requestBody))
+		requestBody, _ := json.Marshal(BatchEventsBody{Batch: payload.Records})
+		req, _ = http.NewRequest("POST", eventsHost+"/v1/events/batch", bytes.NewReader(requestBody))
```
ajwootto, line 141: do we not want to halt execution here or do something with these errors? 

like if it fails to json marshal the current payload we should probably throw it away permanently otherwise we'll be caught in an error loop

eventqueue.go
nsalamad: that's a good question. We should mark it as failed and not retry-able. I'll make the change

eventqueue.go
nsalamad: yeah it's possible that wasm go lib can throw an error https://github.com/DevCycleHQ/go-server-sdk/blob/main/localbucketing.go#L241

eventqueue.go
nsalamad: agreed!

eventqueue.go
```diff
@@ -138,22 +138,45 @@ func (e *EventQueue) FlushEvents() (err error) {
 		var req *http.Request
 		var resp *http.Response
 		requestBody, err := json.Marshal(BatchEventsBody{Batch: payload.Records})
+		if err != nil {
+			log.Printf("Failed to marshal batch events body: %s", err)
+			err = e.localBucketing.onPayloadFailure(payload.PayloadId, false)
```
ajwootto, line 143: I think that marking this failed will cause it to be retried later. If it fails to marshal then the data is probably corrupted and we would need to drop it (in this case by calling `onPayloadSuccess`

nsalamad: the second parameter to `onPayloadFailure` us `retryable`. Passing in `false` will delete it from pending payloads. https://github.com/DevCycleHQ/js-sdks/blob/main/lib/shared/bucketing-assembly-script/assembly/managers/eventQueueManager.ts#L95 and https://github.com/DevCycleHQ/js-sdks/blob/main/lib/shared/bucketing-assembly-script/assembly/eventQueue/requestPayloadManager.ts#L184

ajwootto: oh gotcha.

ajwootto (approved):

PR #55, opened 2023-01-23
await initialization before closing

elliotCamblor:


JamieSinn (approved):

client.go
```diff
@@ -67,7 +68,12 @@ func setLBClient(environmentKey string, options *DVCOptions, c *DVCClient) (*DVC
 
 	if err != nil {
 		if options.OnInitializedChannel != nil {
```
ajwootto, line 70: maybe this whole block should be a separate function? lots of repetition below

PR #58, opened 2023-01-26
fix retryable error logic and swallow initialization errors

ajwootto:
- fix logic for retryable errors to check for 5xx errors

PR #62, opened 2023-01-26
fix issues with close logic 

ajwootto:
- use a buffered channel for the internal initialization system to prevent blocking on writes
- only wait on initialization channel if initialized is false
- change definition of isInitialized to always be true when initialization has finished, even if it was unsuccessful
- change methods like Variable to directly check for whether a config exists rather than checking isInitialized
- increase buffer size of config stop polling channel to prevent problems with calling close when a 403 was received.

### DIFF
--- a/client.go
+++ b/client.go
@@ -67,6 +67,7 @@ func setLBClient(environmentKey string, options *DVCOptions, c *DVCClient) error
 	localBucketing, err := initializeLocalBucketing(environmentKey, options)
 
 	if err != nil {
+		c.isInitialized = true
 		if options.OnInitializedChannel != nil {
 			go func() {
 				options.OnInitializedChannel <- true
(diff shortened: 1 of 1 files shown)


ajwootto:
* main
  * **PR #58** <a href="https://app.graphite.dev/github/pr/DevCycleHQ/go-server-sdk/58" target="_blank">Graphite</a> 
    * **PR #62** <a href="https://app.graphite.dev/github/pr/DevCycleHQ/go-server-sdk/62" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/DevCycleHQ/go-server-sdk/62?utm_source=stack-comment).

client.go
```diff
@@ -114,7 +113,7 @@ func NewDVCClient(environmentKey string, options *DVCOptions) (*DVCClient, error
 	c.DevCycleOptions = options
 
 	if !c.DevCycleOptions.EnableCloudBucketing {
-		c.internalOnInitializedChannel = make(chan bool)
+		c.internalOnInitializedChannel = make(chan bool, 1)
```
elliotCamblor, line 116: I thought that sends to a buffered channel were blocking when the buffer is full, so shouldnt this have a size of 2?

ajwootto: it should only ever be written to once though

elliotCamblor: ah I misunderstood the docs, I thought it would block since the write would fill the buffer

elliotCamblor (approved):

PR #72, opened 2023-02-18
move to global logger

rybit:
Not being able to configure this was making it such that our system had messages that weren't traceable to the service using them, or able to discard them. It was causing a lot of noise in our running system. 

I didn't really do much to look at the contents of the messages, just how they were sending. 

Also, I touched up some errors I saw around ineffectual assignment and some other small errors my linter caught.

### DIFF
--- a/client.go
+++ b/client.go
@@ -6,9 +6,7 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
-	"github.com/matryer/try"
 	"io/ioutil"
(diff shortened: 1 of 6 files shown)


JamieSinn:
Thanks for the PR! 

Looks good to me; will have @ajwootto do a once over and then we can merge

ajwootto (approved):
looks good to me, thanks for the contribution!

PR #77, opened 2023-02-22
DVC-6277 - Create custom logging wrapper

JamieSinn:
Handle for printing, info, warning and error.

We currently have no case where the entire SDK would require a crash of the application; so panic and fatal aren't needed currently.


client.go
```diff
@@ -110,14 +110,17 @@ func NewDVCClient(sdkKey string, options *DVCOptions) (*DVCClient, error) {
 	c.cfg = cfg
 	c.common.client = c
 	c.DevCycleOptions = options
-
+	if c.DevCycleOptions.CustomLogger != nil {
```
ajwootto, line 113: this option should just be called "Logger" to match the Node SDK

client.go
```diff
@@ -441,7 +444,7 @@ func (c *DVCClient) SetClientCustomData(customData map[string]interface{}) error
 			err = c.localBucketing.SetClientCustomData(c.sdkKey, string(data))
 			return err
 		} else {
-			printf("SetClientCustomData called before client initialized")
+			debugf("SetClientCustomData called before client initialized")
```
ajwootto, line 447: all these should be warnings or errors

client.go
```diff
@@ -57,7 +57,7 @@ func initializeLocalBucketing(sdkKey string, options *DVCOptions) (ret *DevCycle
 	ret = &DevCycleLocalBucketing{}
 	err = ret.Initialize(sdkKey, options, cfg)
 	if err != nil {
-		printf("error while initializing local bucketing", err)
+		warnf("error while initializing local bucketing", err)
```
ajwootto, line 60: shouldnt this be an error not a warning?

configuration.go
```diff
@@ -64,15 +63,16 @@ type DVCOptions struct {
 	EventsAPIURI                 string
 	OnInitializedChannel         chan bool
 	BucketingAPIURI              string
+	CustomLogger                 Logger
 }
 
 func (o *DVCOptions) CheckDefaults() {
 	if o.EventFlushIntervalMS < time.Millisecond*500 || o.EventFlushIntervalMS > time.Minute*1 {
-		log.Println("EventFlushIntervalMS cannot be less than 500ms or longer than 1 minute. Defaulting to 30 seconds.")
+		infof("EventFlushIntervalMS cannot be less than 500ms or longer than 1 minute. Defaulting to 30 seconds.")
```
ajwootto, line 71: these should be warnings or errors

eventqueue.go
```diff
@@ -141,10 +136,10 @@ func (e *EventQueue) FlushEvents() (err error) {
 		var resp *http.Response
 		requestBody, err := json.Marshal(BatchEventsBody{Batch: payload.Records})
 		if err != nil {
-			log.Printf("Failed to marshal batch events body: %s", err)
+			warnf("Failed to marshal batch events body: %s", err)
```
ajwootto, line 139: isnt this an error?

eventqueue.go
```diff
@@ -141,10 +136,10 @@ func (e *EventQueue) FlushEvents() (err error) {
...
 			err = e.localBucketing.onPayloadFailure(payload.PayloadId, false)
 			if err != nil {
-				log.Println(err)
+				warnf("Failed to mark payload as failed: %s", err)
```
ajwootto, line 142: error?

eventqueue.go
```diff
@@ -153,7 +148,7 @@ func (e *EventQueue) FlushEvents() (err error) {
 			log.Printf("Failed to create request to events api: %s", err)
 			err = e.localBucketing.onPayloadFailure(payload.PayloadId, false)
 			if err != nil {
-				log.Println(err)
+				warnf("Failed to mark payload as failed: %s", err)
```
ajwootto, line 151: error?

eventqueue.go
```diff
@@ -165,55 +160,54 @@ func (e *EventQueue) FlushEvents() (err error) {
...
 		}
 
 		if resp.StatusCode >= 500 {
-			err = e.localBucketing.onPayloadFailure(payload.PayloadId, true)
-			if err != nil {
-				log.Println(err)
-				continue
-			}
-			log.Println("Server error, retrying later")
+			debugf("Server error, retrying later")
```
ajwootto, line 169: warning?

eventqueue.go
```diff
@@ -165,55 +160,54 @@ func (e *EventQueue) FlushEvents() (err error) {
...
-			log.Println("Error sending events", string(responseBody))
+			warnf("Error sending events - Response: %s", string(responseBody))
 			continue
 		}
 
 		if resp.StatusCode == 201 {
 			err = e.localBucketing.onPayloadSuccess(payload.PayloadId)
 			if err != nil {
-				log.Println(err)
+				warnf("failed to mark payload as success %s", err)
```
ajwootto, line 194: error?

localbucketing.go
```diff
@@ -329,19 +329,17 @@ func (d *DevCycleLocalBucketing) GenerateBucketedConfigForUser(user string) (ret
...
-	}
-	return ret, nil
+	return ret, err
 }
 
 func (d *DevCycleLocalBucketing) StoreConfig(sdkKey, config string) error {
 	defer func() {
 		if err := recover(); err != nil {
-			printf("Failed to process config: ", err)
+			warnf("Failed to process config: ", err)
```
ajwootto, line 342: error?

client.go
JamieSinn: Will swap

client.go
JamieSinn: I wasn't sure between a debug statement or a warning; any specific reason for it being a warning?

ajwootto: it's a user error to call this method before initialization is finished, so it shouldn't be suppressed as a debug log

eventqueue.go
JamieSinn: There's nothing that a user can do with this message; hence hiding it behind debug.

ajwootto: isn't it important to know the network requests aren't succeeding? seems more important than "debug" level to me

eventqueue.go
JamieSinn: This is already dumped as an error via WASM.

client.go
```diff
@@ -110,14 +110,17 @@ func NewDVCClient(sdkKey string, options *DVCOptions) (*DVCClient, error) {
...
+	}
 	if !c.DevCycleOptions.EnableCloudBucketing {
 		c.internalOnInitializedChannel = make(chan bool, 1)
 		if c.DevCycleOptions.OnInitializedChannel != nil {
+			// TODO: Pass this error back via a channel internally
 			go func() {
 				err := setLBClient(sdkKey, options, c)
 				if err != nil {
-					printf(err.Error())
+					infof(err.Error())
```
ajwootto, line 123: should this be an error since its the async analog of returning the err on initialization

client.go
```diff
@@ -238,7 +241,7 @@ func (c *DVCClient) Variable(userdata DVCUser, key string, defaultValue interfac
 			variableEvaluationType = EventType_AggVariableEvaluated
 		} else {
 			if !sameTypeAsDefault && bucketed.Variables[key].Value != nil {
-				printf("Type mismatch for variable %s. Expected type %s, got %s",
+				debugf("Type mismatch for variable %s. Expected type %s, got %s",
```
ajwootto, line 244: warning

client.go
```diff
@@ -163,7 +166,7 @@ func (c *DVCClient) AllFeatures(user DVCUser) (map[string]Feature, error) {
 			user, err := c.generateBucketedConfig(user)
 			return user.Features, err
 		} else {
-			printf("AllFeatures called before client initialized")
+			warnf("AllFeatures called before client initialized")
```
ajwootto, line 169: given that these are implementation errors, I really think they should be error logs

client.go
```diff
@@ -291,7 +294,7 @@ func (c *DVCClient) Variable(userdata DVCUser, key string, defaultValue interfac
 				variable.Value = localVarReturnValue.Value
 				variable.IsDefaulted = false
 			} else {
-				printf("Type mismatch for variable %s. Expected type %s, got %s",
+				debugf("Type mismatch for variable %s. Expected type %s, got %s",
```
ajwootto, line 297: warning

client.go
JamieSinn: There's no way to bubble that error up beyond the log message - hence the ToDo

eventqueue.go
```diff
@@ -36,50 +34,48 @@ type EventQueueOptions struct {
...
+		err = e.localBucketing.initEventQueue(string(eventQueueOpt))
 		ticker := time.NewTicker(e.options.EventFlushIntervalMS)
 
 		go func() {
 			for {
 				select {
 				case <-e.flushStop:
 					ticker.Stop()
-					log.Println("Stopping event flushing.")
+					warnf("Stopping event flushing.")
```
ajwootto, line 58: does this deserve a warning? it will also happen when you call close

JamieSinn: I think it does, given the potential for it to happen when a malformed config response is given (unexpected response code)

ajwootto: then it should be separated from the legitimate one caused by close maybe?

client.go
JamieSinn: But this is not something that someone _has_ to do something to fix. It's optional, and I feel like warning is a better fit here.

client.go
JamieSinn: Once we resolve the JSON/Number issue - yes, but until then that'd just be spammy

client.go
ajwootto: true, I think we use warning in other SDKs actually

eventqueue.go
JamieSinn: yep - but that's outside this scope.

client.go
ajwootto: yeah but it could still print at error level could it not?

eventqueue.go
ajwootto: wait this is the eventqueue one, this can only ever happen when close is called. 
The configmanager one could happen when the config response is wrong, but that one is currently `info`

JamieSinn: will swap the two

client.go
ajwootto: then why print it again here?

JamieSinn: will strip.

eventqueue.go
```diff
@@ -165,55 +153,54 @@ func (e *EventQueue) FlushEvents() (err error) {
...
+			infof("Flushed %d events\n", payload.EventCount)
 		}
 	}
 	return err
 }
 
+func reportPayloadFailure(localBucketing *DevCycleLocalBucketing, payloadId string, retry bool) (err error) {
+	err = localBucketing.onPayloadFailure(payloadId, retry)
+	if err != nil {
+		warnf("Failed to mark payload as failed: %s", err)
```
ajwootto, line 199: this still seems like an error to me, it means something has gone wrong with the internal operation of the SDK and its no longer going to be tracking events correctly

eventqueue.go
```diff
@@ -165,55 +153,54 @@ func (e *EventQueue) FlushEvents() (err error) {
...
+			_ = reportPayloadFailure(e.localBucketing, payload.PayloadId, true)
 			continue
 		}
 
 		if resp.StatusCode >= 400 {
-			err = e.localBucketing.onPayloadFailure(payload.PayloadId, false)
-			if err != nil {
-				log.Println(err)
+			reportError := reportPayloadFailure(e.localBucketing, payload.PayloadId, false)
+			if reportError != nil {
```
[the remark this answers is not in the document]
JamieSinn: Fair - will re-order.

eventqueue.go
JamieSinn: Yeah that makes sense. Swapped

ajwootto (approved):
if you rebase the test harness should pass

PR #80, opened 2023-02-23
slim down some logging

rybit:
I had these deployed in our staging environment and noticed that it was kinda noisy. 

re: flush - this is debug level info imho. 
re: unchanged - there is an existing 'updated the config' statement and error notifications


JamieSinn:
@ajwootto you'll have some opinions here :P

JamieSinn (approved):

PR #84, opened 2023-02-28
fix: log variable defaulted event when calling variable before init

ajwootto:
- if calling `variable` before init, log a `aggVariableDefaulted` event
- change deferred initialization to only defer the actual config HTTP call, but synchronously instantiate everything else.

### DIFF
--- a/client.go
+++ b/client.go
@@ -80,7 +80,13 @@ func setLBClient(sdkKey string, options *DVCOptions, c *DVCClient) error {
 	}
 
 	c.localBucketing = localBucketing
-	c.configManager = c.localBucketing.configManager
+	c.configManager = &EnvironmentConfigManager{localBucketing: localBucketing}
+	err = c.configManager.Initialize(sdkKey, localBucketing)
+
+	if err != nil {
+		return err
+	}
+
 	c.eventQueue = c.localBucketing.eventQueue
 	c.isInitialized = true
 	if options.OnInitializedChannel != nil {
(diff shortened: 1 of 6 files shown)


client.go
```diff
@@ -115,13 +115,17 @@ func NewDVCClient(sdkKey string, options *DVCOptions) (*DVCClient, error) {
...
 		c.internalOnInitializedChannel = make(chan bool, 1)
+		err := setLBClient(sdkKey, options, c)
+		if err != nil {
+			return c, err
+		}
 		if c.DevCycleOptions.OnInitializedChannel != nil {
 			// TODO: Pass this error back via a channel internally
 			go func() {
-				_ = setLBClient(sdkKey, options, c)
+				_ = c.configManager.Initialize(sdkKey, c.localBucketing)
```
JamieSinn: Do you not want the relationship to be doublely linked now?

I guess it wasn't used in reality.

localbucketing_test.go
```diff
@@ -160,34 +160,6 @@ func BenchmarkDevCycleLocalBucketing_GenerateBucketedConfigForUser(b *testing.B)
...
-	err = localBucketing.SetPlatformData(`{"platform": "golang-testing", "sdkType": "server", "platformVersion": "testing", "deviceModel": "testing", "sdkVersion":"testing"}`)
-	if err != nil {
-		t.Fatal(err)
-	}
-	localBucketing.configManager.cancel()
-	localBucketing.configManager.context.Done()
-	fmt.Println("done")
-}
-
 func TestEnvironmentConfigManager_LocalBucketing(t *testing.T) {
```
JamieSinn, line 163: you could delete this one too

PR #90, opened 2023-03-06
Perf improvements on top of dc's

paulo:


JamieSinn:
Closed as the majority of these are integrated.

PR #92, opened 2023-03-06
Byte array reads

ajwootto:
Implement @paulo 's changes to reading strings out of WASM by preallocating the desired memory size and using byte arrays instead of strings.

### DIFF
--- a/localbucketing.go
+++ b/localbucketing.go
@@ -87,6 +87,7 @@ func (d *DevCycleLocalBucketing) Initialize(sdkKey string, options *DVCOptions, 
 	}
 
 	err = d.wasmLinker.DefineFunc(d.wasmStore, "env", "abort", func(messagePtr, filenamePointer, lineNum, colNum int32) {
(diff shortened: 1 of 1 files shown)


localbucketing.go
```diff
@@ -478,18 +483,22 @@ func (d *DevCycleLocalBucketing) newAssemblyScriptString(param string) (int32, e
 // https://www.assemblyscript.org/runtime.html#memory-layout
 // This skips every other index in the resulting array because
 // there isn't a great way to parse UTF-16 cleanly that matches the WTF-16 format that ASC uses.
```
JamieSinn, line 485: since this isn't string specific - do we want to remove the comment to clarify?

JamieSinn (approved):

localbucketing.go
ajwootto: it's still doing these things though, it's just returning a byte array that can be turned into a valid string rather than the string itself

PR #93, opened 2023-03-06
Fix UTF-16 handling

JamieSinn:
Fix allocation to assume no conversion from utf8 is needed - just byte padding.

### DIFF
--- a/localbucketing.go
+++ b/localbucketing.go
@@ -8,7 +8,6 @@ import (
 	"math/rand"
 	"sync"
 	"time"
-	"unicode/utf16"
 	"unsafe"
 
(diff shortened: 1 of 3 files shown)


localbucketing.go
```diff
@@ -190,7 +189,7 @@ func (d *DevCycleLocalBucketing) initEventQueue(options string) (err error) {
 	errorMessage = ""
 	defer d.wasmMutex.Unlock()
 
-	optionsAddr, err := d.newAssemblyScriptString(options)
+	optionsAddr, err := d.newAssemblyScriptString([]byte(options))
```
ajwootto, line 192: I think the `initEventQueue` method could also take a byte array

ajwootto (approved):

PR #89, opened 2023-03-03
chore: add new variableForUser method from WASM

ajwootto:
- use more efficient `variableForUser` call in WASM to get a single variable value

PR #95, opened 2023-03-06
DVC-6496: add benchmark and sanitized config

rob-odwyer:
To run this benchmark:
```
go test -run=^$ -bench ^BenchmarkDVCClient_Variable$ ./
```

Currently I get results like this:
```
goos: darwin
goarch: amd64
pkg: github.com/devcyclehq/go-server-sdk/v2
cpu: Intel(R) Core(TM) i7-9750H CPU @ 2.60GHz
BenchmarkDVCClient_Variable-12    	2023/03/06 15:07:35 INFO: Config set. ETag: TESTING
2023/03/06 15:07:35 INFO: DevCycle SDK Initialized.
2023/03/06 15:07:36 INFO: Config set. ETag: TESTING
2023/03/06 15:07:36 INFO: DevCycle SDK Initialized.
     144	   8544955 ns/op	  170185 B/op	    2069 allocs/op
PASS
ok  	github.com/devcyclehq/go-server-sdk/v2	2.738s
```
But it's significantly faster when run on top of #89.

The large config file is the sanitized one Chris produced from the production config we've been using thus far.


testdata/fixture_small_config.json
```diff
@@ -0,0 +1,99 @@
+{
```
rob-odwyer, line 1: This is just the original sample config used by the tests, pretty-printed and moved to a file for easier diffing in the future.

testing_helpers.go
```diff
@@ -1,15 +1,30 @@
...
+	//go:embed testdata/fixture_large_config.json
+	test_large_config          string
+	test_large_config_variable = "v-key-25"
 )
 
+func init() {
+	// Remove newlines in configs
+	test_config = strings.ReplaceAll(test_config, "\n", "")
+	test_large_config = strings.ReplaceAll(test_large_config, "\n", "")
+}
```
rob-odwyer, line 26: Not sure if removing the newlines is actually needed - the tests seemed to pass either way.

ajwootto (approved):

PR #106, opened 2023-03-08
Memory allocation pool

ajwootto:
adapted from @paulo 's memory allocation pool.

Rather than using sync pools and trying to generically allocate new blocks within that pool, we assume there will only ever be at most two allocated blocks needed per "bucket" of memory size. 

If the current string is too large to fit in an existing bucket, it will dynamically allocate the memory for that string and log a warning to indicate a potential performance problem.

Make the number of buckets controllable with an option. The pool can also be turned off entirely.

**Note** After the protobuf changes, we will only need one block per bucket because there will only ever be one contiguous buffer of data being passed into the WASM methods (Containing all the arguments in a structured object)


configuration.go
```diff
@@ -49,6 +49,13 @@ type APIKey struct {
 	Prefix string
 }
 
+type AdvancedOptions struct {
+	// controls the maximum number of pre-allocated memory blocks used for WASM execution. This influences the maximum
+	// string length that can be fit inside of preallocated memory
+	// Can be set to -1 to disable pre-allocated memory blocks entirely.
+	MaxMemoryAllocationBuckets int
```
JamieSinn, line 56: This should be noted that it's $\sum_{n=1}^n 2^n$ memory usage

localbucketing.go
```diff
@@ -133,7 +137,7 @@ func (d *DevCycleLocalBucketing) Initialize(wasmMain *WASMMain, sdkKey string, o
 	d.setPlatformDataFunc = d.wasmInstance.GetExport(d.wasmStore, "setPlatformData").Func()
 	d.setClientCustomDataFunc = d.wasmInstance.GetExport(d.wasmStore, "setClientCustomData").Func()
 	d.setConfigDataFunc = d.wasmInstance.GetExport(d.wasmStore, "setConfigData").Func()
-	d.variableForUserFunc = d.wasmInstance.GetExport(d.wasmStore, "variableForUser").Func()
+	d.variableForUserFunc = d.wasmInstance.GetExport(d.wasmStore, "variableForUserPreallocated").Func()
```
JamieSinn, line 140: can we keep the other in for benchmarking purposes?

ajwootto: why? then we'd have to maintain another code path for it etc

JamieSinn: mostly to prevent us having to run two different code checkouts to test perf.

ajwootto: that's the same as all the other changes we've made so far. I think it would get unwieldly to hold on to older, slower code

JamieSinn: yeah for sure - all good

configuration.go
```diff
@@ -49,6 +49,13 @@ type APIKey struct {
 	Prefix string
 }
 
+type AdvancedOptions struct {
+	// controls the maximum number of pre-allocated memory blocks used for WASM execution. This influences the maximum
+	// string length that can be fit inside of preallocated memory
+	// Can be set to -1 to disable pre-allocated memory blocks entirely.
```
JamieSinn, line 55:
```suggestion
	// Can be set to -1 to disable pre-allocated memory blocks entirely.
	// This takes \sum_{n=1}^{n} 2^n memory usage
```

ajwootto: is this way of writing it parsed by some kind of thing to give a readable formula? If not there's probably a simpler way to convey this information

JamieSinn: That's LaTeX - pretty standard imo

configuration.go
ajwootto: it's actually twice that amount since there are two blocks being allocated for each K

JamieSinn: I was expecting we remove the 2x for the protobuf change, but yes

ajwootto: right

localbucketing.go
```diff
@@ -153,6 +157,27 @@ func (d *DevCycleLocalBucketing) Initialize(wasmMain *WASMMain, sdkKey string, o
 		JSON:    VariableTypeCode(jsonType),
 	}
 
+	d.allocatedMemPool = make([][]int32, d.options.MaxMemoryAllocationBuckets) // overly simplistic, probably need a larger pool
+
+	for i := 0; i < d.options.MaxMemoryAllocationBuckets; i++ {
```
paulo, line 162: Not a blocker, but in the future it's probably not worth allocating strings with a length of 1

localbucketing.go
```diff
@@ -532,6 +564,60 @@ func (d *DevCycleLocalBucketing) newAssemblyScriptString(param []byte) (int32, e
...
+
+	return ptr, preAllocated, nil
+}
+
+func (d *DevCycleLocalBucketing) allocMemForStringPool(size int32, preAllocIndex int32) (addr int32, preAllocated bool, err error) {
+	cachedIdx := int32(math.Ceil(math.Log2(float64(size))))
+	// index is the highest power value of 2 for the size we want
+
+	// if this length exceeds the max size of the pool, we'll just allocate the memory temporarily
+	if len(d.allocatedMemPool) == 0 {
```
paulo, line 591: nit: this check should be used as a guard before `cachedIdx` is calculated

localbucketing.go
ajwootto: yeah I was thinking that as well. Just pushed a change that starts the blocks at 32 bytes instead.

localbucketing.go
```diff
@@ -532,6 +575,61 @@ func (d *DevCycleLocalBucketing) newAssemblyScriptString(param []byte) (int32, e
...
+func (d *DevCycleLocalBucketing) allocMemForString(size int32) (addr int32, err error) {
+	const objectIdString int32 = 2
+
+	// malloc
+	ptr, err := d.__newFunc.Call(d.wasmStore, size, objectIdString)
+	if err != nil {
+		return -1, err
+	}
+
+	if err := d.assemblyScriptPin(ptr.(int32)); err != nil {
```
JamieSinn, line 626: Do we not care about unpinning this memory during close/etc? Maybe a future thing because of the multi-sdk-key usage format.

JamieSinn (approved):

localbucketing.go
ajwootto: yeah there's a bunch of work we need to do to be able to close cleanly, we can note that as part of that work I think

PR #108, opened 2023-03-08
HTTP Benchmarking - pre-threads

JamieSinn:


bench/benchy.go
```diff
@@ -0,0 +1,34 @@
...
+	"strconv"
+
+	"github.com/gin-gonic/gin"
+)
+
+func main() {
+
+	variableKey := os.Getenv("DVC_VARIABLE_KEY")
+	userId := os.Getenv("DVC_USER_ID")
+	client, err := devcycle.NewDVCClient(os.Getenv("DVC_SERVER_KEY"), &devcycle.DVCOptions{})
```
ajwootto, line 15: lets use `DVC_SERVER_SDK_KEY` consistently everywhere

JamieSinn: will swap - but need to switch this away from a full go module to just a file - it tries to source the public version instead of the local one

ajwootto: bump

JamieSinn: right - didn't copy that - sec

ajwootto (approved):

ajwootto (changes requested):
dependency question

go.mod
```diff
@@ -4,8 +4,34 @@ go 1.19
 
 require (
 	github.com/bytecodealliance/wasmtime-go/v6 v6.0.0
+	github.com/gin-gonic/gin v1.9.0
```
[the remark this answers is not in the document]
JamieSinn: Not intended to be merged in this format - we're going to change this to pull from released SDK versions instead.

This version just allows you to do local benching.

PR #102, opened 2023-03-07
add support for managed thread pool

ajwootto:

PR #103, opened 2023-03-07
refactor wasm code to split reusable stuff from per-thread stuff

ajwootto:

PR #109, opened 2023-03-08
DVC-6495: thread pool with event flushing

rob-odwyer:
This adds event handling on top of #102.

To run the concurrent benchmark with events enabled:
```
go test -run=^$ -bench ^BenchmarkDVCClient_VariableConcurrent$ -benchEnableEvents ./
```


thread_pool.go
```diff
@@ -24,22 +31,72 @@ type VariableForUserResponse struct {
...
+	eventQueueOpt, err = json.Marshal(options.eventQueueOptions())
+	if err != nil {
+		return fmt.Errorf("error marshalling event queue options: %w", err)
+	}
+	err = w.localBucketing.initEventQueue(eventQueueOpt)
+	if err != nil {
+		return fmt.Errorf("error initializing worker event queue: %w", err)
+	}
+
+	ticker := time.NewTicker(options.EventFlushIntervalMS)
```
ajwootto, line 51: should we make sure these are offset from each other so that we dont have all of the workers flushing at the same time and locking up the pool from processing variables?

eventqueue.go
```diff
@@ -153,47 +150,64 @@ func (e *EventQueue) FlushEvents() (err error) {
...
-			debugf("Flushed %d events\n", payload.EventCount)
 		}
 	}
 	return err
 }
 
-func reportPayloadFailure(localBucketing *DevCycleLocalBucketing, payloadId string, retry bool) (err error) {
-	err = localBucketing.onPayloadFailure(payloadId, retry)
+func (e *EventQueue) reportPayloadSuccess(payload FlushPayload, useChannel bool) (err error) {
+	if useChannel {
```
ajwootto, line 190: could all this be simplified by just having the eventqueue ask for events from the worker rather than passively receiving them on the channel? it could use the same external call mechanism as the StoreConfig method

it would let you only call the `onPayloadSuccess` when the messages have actually been flushed, rather than having to worry about requeueing them (which is more of the intent behind the current interface for the event queue in WASM)

thread_pool.go
rob-odwyer: yeah, I was thinking about adding some kind of jitter too. Think adding a random duration to the interval is enough, or should it be based on something like the worker index?

ajwootto: might as well base it on the worker index so we have a predictable separation rather than risk two workers colliding due to random chance

ajwootto:
* main
  * **PR #103** <a href="https://app.graphite.dev/github/pr/DevCycleHQ/go-server-sdk/103" target="_blank">Graphite</a> 
    * **PR #102** <a href="https://app.graphite.dev/github/pr/DevCycleHQ/go-server-sdk/102" target="_blank">Graphite</a> 
      * **PR #109** <a href="https://app.graphite.dev/github/pr/DevCycleHQ/go-server-sdk/109" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/DevCycleHQ/go-server-sdk/109?utm_source=stack-comment).

ajwootto:
* main
  * **PR #103** <a href="https://app.graphite.dev/github/pr/DevCycleHQ/go-server-sdk/103" target="_blank">Graphite</a> 
    * **PR #102** <a href="https://app.graphite.dev/github/pr/DevCycleHQ/go-server-sdk/102" target="_blank">Graphite</a> 
      * **PR #109** <a href="https://app.graphite.dev/github/pr/DevCycleHQ/go-server-sdk/109" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/DevCycleHQ/go-server-sdk/109?utm_source=stack-comment).

PR #116, opened 2023-03-13
fix local bucketing unpinning buffer data

ajwootto:
- fix behaviour when reserved memory blocks are too small and new memory is dynamically allocated
- previously, any dynamic allocation would attempt to unpin the ptr representing its data, however this ptr actually pointed to a permanently reserved block of memory for storing header information about the separately-reserved byte array
- it was then incorrectly unpinning this header, which made it liable to be garbage collected, thus breaking the ability to pass byte buffers into WASM

### DIFF
--- a/localbucketing_test.go
+++ b/localbucketing_test.go
@@ -3,6 +3,7 @@ package devcycle
 import (
 	_ "embed"
 	"fmt"
+	"github.com/devcyclehq/go-server-sdk/v2/proto"
 	"github.com/jarcoal/httpmock"
 	"testing"
 )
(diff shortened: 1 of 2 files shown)


localbucketing_test.go
```diff
@@ -162,3 +163,70 @@ func BenchmarkDevCycleLocalBucketing_GenerateBucketedConfigForUser(b *testing.B)
...
+	if err != nil {
+		b.Fatal(err)
+	}
+	_, err = localBucketing.VariableForUser_PB(paramsBuffer)
+
+	if err != nil {
+		b.Fatal(err)
+	}
+
+	err = localBucketing.assemblyScriptCollect()
```
chris-hoefgen, line 221: So effectively the test is:

Get a variable
Run garbage collection
try and get another variable

chris-hoefgen (approved):

PR #129, opened 2023-03-17
POC of using object pool for variable evaluations

ajwootto:


bucketing_pool.go
```diff
@@ -0,0 +1,171 @@
...
+	}
+
+	//start := time.Now()
+
+	processAll := func(processPool *pool.ObjectPool) error {
+		i := 0
+		curObjects := make([]*BucketingPoolObject, processPool.Config.MaxTotal)
+		for i < processPool.Config.MaxTotal {
+			borrowed, err := processPool.BorrowObject(p.ctx)
+			curObj := borrowed.(*BucketingPoolObject)
```
rob-odwyer, line 107: I think this is currently safe given our pool config, but just for future safety - we should check for the error before casting in case `nil, <error>` is returned.

bucketing_pool.go
```diff
@@ -0,0 +1,171 @@
...
+			if err != nil {
+				_ = processPool.ReturnObject(p.ctx, curObj)
```
rob-odwyer, line 109: Is returning it when `BorrowObject` returns an error the right thing to do? Shouldn't we assume we didn't succeed in borrowing anything?

rob-odwyer: Although if `ReturnObject` is just going to give us an error we can ignore, maybe it's safe? We might be passing a nil in here though in this case.

bucketing_pool.go
```diff
@@ -0,0 +1,171 @@
...
+	//debugf("Borrowed all objects from pool for %s in %s", operationName, time.Since(start))
+
+	return err
+}
+
+func (p *BucketingPool) SetConfig(config []byte) error {
+	if p.closed {
+		return errorf("Cannot set config on closed pool")
+	}
+	p.configData = &config
```
rob-odwyer, line 149: Do we need to save a copy of configData here? Same with customData below.

bucketing_pool_factory.go
```diff
@@ -0,0 +1,51 @@
...
+	var bucketing = &BucketingPoolObject{}
+	err := bucketing.Initialize(f.wasmMain, f.sdkKey, f.options)
+	if err != nil {
+		return nil, err
+	}
+
+	return pool.NewPooledObject(
+			bucketing),
+		nil
+}
```
rob-odwyer, line 34: We could simplify this a bit by moving these methods straight into `BucketingPool` so it implements the factory interface. Not required right now though.

bucketing_pool_object.go
```diff
@@ -0,0 +1,89 @@
...
+	"sync/atomic"
+)
+
+var (
+	objectId = atomic.Int32{}
+)
+
+type BucketingPoolObject struct {
+	localBucketing   *DevCycleLocalBucketing
+	flushResultChan  chan *FlushResult
```
rob-odwyer, line 15: unused channel

bucketing_pool_object.go
```diff
@@ -0,0 +1,89 @@
...
+	id               int32
+	configData       *[]byte
+	clientCustomData *[]byte
+	lastFlushTime    int64
+}
+
+func (o *BucketingPoolObject) Initialize(wasmMain *WASMMain, sdkKey string, options *DVCOptions) (err error) {
+	o.localBucketing = &DevCycleLocalBucketing{}
+	err = o.localBucketing.Initialize(wasmMain, sdkKey, options)
```
rob-odwyer, line 24: We're not returning if error is set here.

client.go
```diff
@@ -270,23 +268,7 @@ func (c *DVCClient) variableForUserProtobuf(user DVCUser, key string, variableTy
...
-			VariableEvalParams: &paramsBuffer,
-		})
-
-		var variableResult = result.(WorkerPoolResponse)
-		if variableResult.Err != nil {
-			return Variable{}, variableResult.Err
-		}
-		variablePB = variableResult.Variable
-	}
+	variablePB, err := c.bucketingObjectPool.VariableForUser(paramsBuffer)
```
rob-odwyer, line 271: It looks like we don't check the error here either? We're checking for the variable being nil, but we should probably do something with that error, or any bug reports we get are going to be pretty light on detail.

eventqueue.go
```diff
@@ -132,32 +138,29 @@ func (e *EventQueue) FlushEvents() (err error) {
...
+
+		for {
+			select {
+			case result := <-respChan:
+				err = object.HandleFlushResults(result)
+				return nil
 			}
 		}
-	}
+	})
```
rob-odwyer, line 161: Not blocking this PR, but we could look at passing a reference to the BucketingPoolObject (or internal local bucketing) through the event queue methods used here. Since it's synchronous now, we could skip using the channel entirely.

rob-odwyer (commented):
Looks good overall, just found a few places with missing error handling.

bucketing_pool.go
```diff
@@ -0,0 +1,171 @@
...
+	"sync"
+	"sync/atomic"
+)
+
+type BucketingPool struct {
+	pool1            *pool.ObjectPool
+	pool2            *pool.ObjectPool
+	currentPool      atomic.Pointer[pool.ObjectPool]
+	poolSwapMutex    sync.Mutex
+	ctx              context.Context
```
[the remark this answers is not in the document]
ajwootto: yeah our use of contexts isn't really correct.

My understanding is that you'd want the client to hold a context that is tied to the overall lifecycle of the app or something, and have the client thread that context automatically through everything else so we can cancel / cleanup stuff when the context is closed?

bucketing_pool.go
ajwootto: this is left over from the original way I was doing the updating of all the workers. It isn't necessary at the moment so I could remove it

bucketing_pool_factory.go
ajwootto: yeah we actually dont need this factory at all anymore because we aren't overriding any of the lifecycle methods, but I might save that for a future cleanup

eventqueue.go
ajwootto: yeah definitely, I tried to fit this into the old channel approach to minimize rewriting but we don't really need to do that

bucketing_pool.go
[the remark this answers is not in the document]
rob-odwyer: Yeah - it can be the app lifecycle, or a HTTP request (net/http requests have them by default and cancel the context if the client disconnects).

rob-odwyer (approved):

PR #132, opened 2023-03-22
stop using channel to handle flush results and unify result handling

ajwootto:
- remove channel callback system for receiving event flush results, and return them directly
- handle all results using a single mutex
- handle results the same way for both the global and pooled local bucketing instances

### DIFF
--- a/eventqueue.go
+++ b/eventqueue.go
@@ -135,11 +135,13 @@ func (e *EventQueue) FlushEvents() (err error) {
 	}
 	e.eventsFlushed.Add(int32(len(payloads)))
 
-	err = e.flushEventPayloads(&PayloadsAndChannel{payloads: payloads})
+	result, err := e.flushEventPayloads(payloads)
 
 	if err != nil {
 		return
 	}
+
+	e.localBucketing.HandleFlushResults(result)
 
 	err = e.bucketingObjectPool.ProcessAll("FlushEvents", func(object *BucketingPoolObject) error {
 		payloads, err := object.FlushEvents()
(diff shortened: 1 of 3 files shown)


eventqueue.go
```diff
@@ -237,33 +233,24 @@ func (e *EventQueue) flushEventPayload(
...
-	var failures []string
-	var retryableFailures []string
-
-	if payloadsAndChannel.channel != nil {
-		successes = make([]string, 0)
-		failures = make([]string, 0)
-		retryableFailures = make([]string, 0)
-	}
+func (e *EventQueue) flushEventPayloads(payloads []FlushPayload) (result *FlushResult, err error) {
+	e.eventsFlushed.Add(int32(len(payloads)))
```
JamieSinn, line 240: Given we use int32 for all params to wasm, I wonder what the performance implications are of this every time. Probably nothing?

localbucketing.go
```diff
@@ -507,6 +516,31 @@ func (d *DevCycleLocalBucketing) SetClientCustomData(customData []byte) error {
 	return err
 }
 
+func (d *DevCycleLocalBucketing) HandleFlushResults(result *FlushResult) {
+	d.wasmMutex.Lock()
+	d.errorMessage = ""
+	defer d.wasmMutex.Unlock()
+
+	for _, payloadId := range result.SuccessPayloads {
```
JamieSinn, line 524: Why not do this as a range of all, and switch case inside?

eventqueue.go
JamieSinn: Was more of a shower thought about how were casting all ints from len to int32 in all the wasm calls may have performance costs?

Mostly just an idle comment for note later.

localbucketing.go
ajwootto: yeah but I'm a human who has to read this later

JamieSinn (approved):

PR #133, opened 2023-03-22
additional safety around constructing new objects for pool

ajwootto:
Update bucketing object factory to make sure that if AddObject is called sometime after initialization, the object that is created will be initialized with the last known config and client custom data. 

AddObject currently shouldn't be called after initialization, but the pool will automatically call it if an object is destroyed or removed from the pool for any reason.

### DIFF
--- a/bucketing_pool.go
+++ b/bucketing_pool.go
@@ -9,13 +9,15 @@ import (
 )
 
 type BucketingPool struct {
-	pool1         *pool.ObjectPool
-	pool2         *pool.ObjectPool
-	currentPool   atomic.Pointer[pool.ObjectPool]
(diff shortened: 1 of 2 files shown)


bucketing_pool.go
```diff
@@ -142,6 +144,7 @@ func (p *BucketingPool) SetConfig(config []byte) error {
 		return errorf("Cannot set config on closed pool")
 	}
 	debugf("Setting config on all workers")
+	p.configData = &config
```
JamieSinn, line 147: You're taking a pass by value and converting to pass by ref? Maybe make the param a pointer to make it a cleaner chain of custody for the value?

PR #135, opened 2023-03-22
DVC-6731 set json config data with utf8 byte arrays

chris-hoefgen:
* Changed the setConfigData(), setPlatformData(), setClientCustomData() to use the new Uint8Array versions in AssemblyScript to send proper utf-8 encoded string data into the WASM.  

* Adds a new assemblyScript util function to set Uint8Arrays outside of memory pools

* New unit test and unit test data file that contains special characters to test encoding


client_test.go
```diff
@@ -44,6 +45,21 @@ func TestDVCClient_AllVariablesLocal(t *testing.T) {
...
+	defer httpmock.DeactivateAndReset()
+	httpCustomConfigMock(test_environmentKey, 200, test_config_special_characters_var)
+	c, err := NewDVCClient("dvc_server_token_hash", &DVCOptions{})
+	fatalErr(t, err)
+
+	variables, err := c.AllVariables(
+		DVCUser{UserId: "j_test", DeviceModel: "testing"})
+	fatalErr(t, err)
+
+	fmt.Println(variables)
```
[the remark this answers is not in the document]
chris-hoefgen: yes...yes it should

localbucketing.go
```diff
@@ -50,19 +51,23 @@ type DevCycleLocalBucketing struct {
...
-	queueAggregateEventFunc           *wasmtime.Func
-	setClientCustomDataFunc           *wasmtime.Func
-	variableForUserFunc               *wasmtime.Func
-	variableForUser_PBFunc            *wasmtime.Func
+	flushEventQueueFunc                   *wasmtime.Func
+	eventQueueSizeFunc                    *wasmtime.Func
+	onPayloadSuccessFunc                  *wasmtime.Func
+	queueEventFunc                        *wasmtime.Func
+	onPayloadFailureFunc                  *wasmtime.Func
+	generateBucketedConfigForUserFunc     *wasmtime.Func
```
ajwootto, line 59: maybe worth removing the old functions that we aren't using now?

localbucketing.go
```diff
@@ -566,6 +576,67 @@ func (d *DevCycleLocalBucketing) newAssemblyScriptString(param []byte) (int32, e
 	return ptr.(int32), nil
 }
 
+func (d *DevCycleLocalBucketing) newAssemblyScriptUTF16StringNoPoolByteArray(param string) (int32, error) {
+	utf16Output := utf16.Encode(([]rune)(param))
+	var rawBytes []byte
+	for _, i := range utf16Output {
+		rawBytes = binary.LittleEndian.AppendUint16(rawBytes, i)
```
ajwootto, line 583: not a big deal but couldn't you just rewrite onto the same utf16Output buffer here?

localbucketing.go
chris-hoefgen: yeah, there is a Go SDK cleanup task in JIRA to look into that. I just figured we should do that in a separate PR

client_test.go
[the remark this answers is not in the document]
chris-hoefgen: Added some actual validation

ajwootto (approved):

PR #137, opened 2023-03-28
DVC-6781: new package and build for native bucketing

rob-odwyer:
- add new native_bucketing package and files for alternate client implementations
- Run unit tests on native bucketing build

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -19,3 +19,6 @@ jobs:
 
     - name: Test
       run: go test -v ./...
+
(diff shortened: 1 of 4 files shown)


JamieSinn (approved):

rob-odwyer:
It shouldn't break tests, unless I've missed something? Right now passing the `native_bucketing` build tag doesn't change anything.

rob-odwyer:
Also - any idea why those gitStream.cm tests always fail for me? I can't see any extra info on what's failing.

PR #138, opened 2023-03-28
DVC-6790 - Porting ASC Types 

JamieSinn:


go.mod
```diff
@@ -4,10 +4,22 @@ go 1.19
...
-require github.com/cheekybits/is v0.0.0-20150225183255-68e9c0620927 // indirect
+require (
+	github.com/cheekybits/is v0.0.0-20150225183255-68e9c0620927 // indirect
+	github.com/go-playground/locales v0.14.1 // indirect
+	github.com/go-playground/universal-translator v0.18.1 // indirect
+	github.com/leodido/go-urn v1.2.2 // indirect
+	golang.org/x/crypto v0.7.0 // indirect
+	golang.org/x/sys v0.6.0 // indirect
+	golang.org/x/text v0.8.0 // indirect
+)
```
rob-odwyer, line 25: I don't think having a go.mod works once we need to start importing it from the root package. I get this from `go build`:
```
client.go:20:2: no required module provides package github.com/devcyclehq/go-server-sdk/v2/native_bucketing; to add it:
	go get github.com/devcyclehq/go-server-sdk/v2/native_bucketing
```
Removing `go.mod` and `go.sum` fixes it.
If there's a way around this though, I'm happy to have two go.mods.

JamieSinn: That's because it's not merged to main/isn't available publicly yet, or it's not in your GOPATH

rob-odwyer: Sure - if it was already merged, we could reference it. But then we'd have to push changes to this package ahead of changes in the root package using it.

In the `proto` sub-package that doesn't have it's own go.mod, we can just make changes without any extra steps.

rob-odwyer: What's the advantage we get to having a separate go.mod?

JamieSinn: Separate versioned updates, better granularity, and the ability to package externally for other things if needed

robbles (approved):

rob-odwyer (approved):

PR #141, opened 2023-03-29
Removing old WASM function references

chris-hoefgen:
cleaning up some of the WASM functions that were replaced by new versions for the protobuf and UTF8 encoding changes

### DIFF
--- a/localbucketing.go
+++ b/localbucketing.go
@@ -55,14 +55,9 @@ type DevCycleLocalBucketing struct {
 	onPayloadSuccessFunc                  *wasmtime.Func
 	queueEventFunc                        *wasmtime.Func
 	onPayloadFailureFunc                  *wasmtime.Func
-	generateBucketedConfigForUserFunc     *wasmtime.Func
-	setPlatformDataFunc                   *wasmtime.Func
-	setConfigDataFunc                     *wasmtime.Func
 	initEventQueueFunc                    *wasmtime.Func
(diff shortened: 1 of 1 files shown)


localbucketing.go
```diff
@@ -445,7 +435,7 @@ func (d *DevCycleLocalBucketing) VariableForUser_PB(serializedParams []byte) (*p
 		return nil, errorf("Error allocating WASM string: %w", err)
 	}
 
-	varPtr, err := d.variableForUser_PBFunc.Call(d.wasmStore, paramsAddr, int32(len(serializedParams)))
+	varPtr, err := d.variableForUser_PB_Preallocated.Call(d.wasmStore, paramsAddr, int32(len(serializedParams)))
```
JamieSinn, line 438: Are we sure we're switching to this one? 
Otherwise should be fine.

chris-hoefgen: I just renamed the Go reference so that it matched the actual function name from the WASM for consistency.  The actually function name is still the same as it was before:

```
d.variableForUser_PB_Preallocated = d.wasmInstance.GetExport(d.wasmStore, "variableForUser_PB_Preallocated").Func()
```

JamieSinn: Ah good - the inline diff made that confusing whether it was replacing it nor not.

JamieSinn (approved):

PR #142, opened 2023-03-29
DVC-6813: Retry config fetching on network error

rob-odwyer:
This PR restructures EnvironmentConfigManager to make it a bit easier to test, then expands the criteria under which we retry a config request.

Previously we only retried once for 500-level HTTP status codes, now we also retry on connection error, DNS error, etc. and unexpected HTTP statuses. When the retries are exhausted, we still return the error to the client, unless it's a 500 status.

### DIFF
--- a/configmanager.go
+++ b/configmanager.go
@@ -10,64 +10,75 @@ import (
 	"time"
 )
 
+const CONFIG_RETRIES = 1
+
+type ConfigReceiver interface {
+	StoreConfig([]byte) error
+}
+
 type EnvironmentConfigManager struct {
 	sdkKey              string
 	configETag          string
-	localBucketing      *DevCycleLocalBucketing
-	bucketingObjectPool *BucketingPool
+	localBucketing      ConfigReceiver
+	bucketingObjectPool ConfigReceiver
 	firstLoad           bool
 	context             context.Context
-	cancel              context.CancelFunc
+	stopPolling         context.CancelFunc
 	httpClient          *http.Client
 	cfg                 *HTTPConfiguration
 	hasConfig           atomic.Bool
-	pollingStop         chan bool
 	ticker              *time.Ticker
 }
 
-func (e *EnvironmentConfigManager) Initialize(
+func NewEnvironmentConfigManager(
 	sdkKey string,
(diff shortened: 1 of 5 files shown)


configmanager.go
```diff
@@ -86,31 +101,29 @@ func (e *EnvironmentConfigManager) fetchConfig(retrying bool) error {
...
-		e.pollingStop <- true
+		e.stopPolling()
 		return errorf("invalid SDK key. Aborting config polling")
 	case statusCode >= 500:
 		// Retryable Errors. Continue polling.
-		if !retrying {
-			warnf("Retrying config fetch. Status:" + resp.Status)
-			return e.fetchConfig(true)
-		}
 		warnf("Config fetch failed. Status:" + resp.Status)
```
rob-odwyer, line 112: Do we want to return an error when we hit too many 500s? Or just keep polling in the background and returning default values?

chris-hoefgen (approved):
Changes make sense to me and i like the refactoring. Probably worth getting a second pair of eyes on it.

configmanager.go
```diff
@@ -10,64 +10,75 @@ import (
...
 		}
 	}()
-	return
 }
 
-func (e *EnvironmentConfigManager) initialFetch() (err error) {
-	err = e.fetchConfig(false)
-	return
+func (e *EnvironmentConfigManager) initialFetch() error {
+	return e.fetchConfig(CONFIG_RETRIES)
```
ajwootto, line 78: returning the error here is going to make the SDK behave differently than the other SDKs, I think this will fail the test harness

rob-odwyer: On second glance, I think this was always returning the error?
It looks like it's ignoring it, but it's assigning the named return value and then returning.

ajwootto: oh right

configmanager.go
rob-odwyer: I tried to preserve the old behaviour of continuing to poll in the background on 500, even after we've exhausted the initial retries, and NOT returning the error.

I'm not sure that's a good idea though, if we've actually broken something it'll just leave the SDK silently defaulting all variables.

PR #145, opened 2023-03-31
DVC-6664 - validate client options

chris-hoefgen:
Added validation to the options to throw an error when combinations of options will result in the client being in an unsafe state. In this case if you set an OnInitialized channel and EnableCloudBucketing=true, then listen on the channel it will hang indefinitely

### DIFF
--- a/client.go
+++ b/client.go
@@ -124,11 +124,6 @@ func NewDVCClient(sdkKey string, options *DVCOptions) (*DVCClient, error) {
 
 	options.CheckDefaults()
 
-	err := options.Validate()
-	if err != nil {
-		return nil, err
-	}
-
 	c := &DVCClient{sdkKey: sdkKey}
 	c.cfg = cfg
 	c.ctx = context.Background()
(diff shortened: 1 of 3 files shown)


client.go
```diff
@@ -152,6 +152,10 @@ func NewDVCClient(sdkKey string, options *DVCOptions) (*DVCClient, error) {
 			c.handleInitialization()
 			return c, err
 		}
+	} else if c.DevCycleOptions.OnInitializedChannel != nil && c.DevCycleOptions.EnableCloudBucketing {
```
rob-odwyer, line 155: The `&& c.DevCycleOptions.EnableCloudBucketing` is redundant here, right?

client_test.go
```diff
@@ -336,6 +336,39 @@ func TestProduction_Local(t *testing.T) {
...
+
+	dvcOptions = DVCOptions{OnInitializedChannel: nil, EnableCloudBucketing: true}
+	_, err = NewDVCClient(test_environmentKey, &dvcOptions)
+	fatalErr(t, err)
+
+	dvcOptions = DVCOptions{OnInitializedChannel: nil, EnableCloudBucketing: false}
+	_, err = NewDVCClient(test_environmentKey, &dvcOptions)
+	fatalErr(t, err)
+}
+
```
rob-odwyer, line 371: I think we should probably check `isInitialized` and `hasConfig` for the two cases where `EnableCloudBucketing = true`.

chris-hoefgen: okay added new checks for isInitialized.  I think hasConfig() requires a bigger discussion

rob-odwyer: Just realized my typo - I meant checking both when `EnableCloudBucketing` is **false**

rob-odwyer (approved):

PR #148, opened 2023-04-03
DVC-6674 cleanup protobuf wasm workaround

chris-hoefgen:
- fixes a workaround previously implemented in the WASM due to a library bug
- regenerate the protobuf code


rob-odwyer (approved):
Might be good to run a really quick benchmark on both main and this branch to confirm the performance is unaffected after removing this.

PR #149, opened 2023-04-04
DVC-6790: Config parsing, bucketing, and segmentation

rob-odwyer:
Combined PR which includes the config parsing, bucketing, segmentation, and tests for all of them.

I think we are still missing some of the newer optimized methods in the AS, like `generateBucketedVariableForUser`.


native_bucketing/model_filters.go
```diff
@@ -1,86 +1,299 @@
...
+
+func (f AudienceMatchFilter) Type() string {
+	return TypeAudienceMatch
+}
+
+func (f AudienceMatchFilter) Validate() error {
+	return nil
+}
+
+func checkCustomData(data map[string]interface{}, clientCustomData map[string]interface{}, filter CustomDataFilter) bool {
```
JamieSinn, line 254: Since this a readonly operation - I wonder if it's worth attaching as a function off the customdatafilter struct?

rob-odwyer: Follow up on this: thinking about refactoring the internals of the operators and filters so that instead of building a tree and then navigating through it and using type asserts to call different filters/operators, we delegate to the filters themselves.
e.g. replace `_evaluateOperator` and `doesUserPassFilter` with:
```
type FilterOrOperator interface {
    DoesUserPass(audiences map[string]NoIdAudience, user DVCPopulatedUser, clientCustomData map[string]interface{}) bool
}
```

I think it's more important for now to get to a working E2E of a variable call though, after which we can come back to refactor this.

native_bucketing/segmentation.go
```diff
@@ -75,48 +73,31 @@ func filterForAudienceMatch(filter AudienceMatchFilter, configAudiences map[stri
...
 	} else if subType == "deviceModel" {
-		return checkStringsFilter(user.DeviceModel, filter.(UserFilter))
+		return checkStringsFilter(user.DeviceModel, filter.(*UserFilter))
 	} else if subType == "platform" {
-		return checkStringsFilter(user.Platform, filter.(UserFilter))
+		return checkStringsFilter(user.Platform, filter.(*UserFilter))
 	} else if subType == "customData" {
-		if err := filter.(CustomDataFilter).Validate(); err != nil {
+		customDataFilter, ok := filter.(*CustomDataFilter)
+		if !ok {
```
JamieSinn, line 94: could combine this into the next statement

native_bucketing/model_dvc_user.go
```diff
@@ -35,7 +36,7 @@ type DVCPopulatedUser struct {
 }
 
 func (p *DVCPopulatedUser) CombinedCustomData() map[string]interface{} {
-	var ret map[string]interface{}
+	ret := make(map[string]interface{}, len(p.CustomData)+len(p.PrivateCustomData))
```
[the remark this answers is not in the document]
rob-odwyer: It might simplify things, but I think it's also a bit unintuitive? This is the user's combined data, whereas the client custom data is associated with the client.

PR #150, opened 2023-04-04
DVC-6656 Add benchmark GitHub action

chris-hoefgen:
A new action that will run the benchmark tests for variables and use the 3rd party `github-action-benchmark` action compare benchmarks against the previous cached version and report errors when the variable performance is off by 25% or higher

The plugin sets the benchmark results as comments on the commit that trigger the warning/error.

### DIFF
--- a/.github/workflows/bench.yml
+++ b/.github/workflows/bench.yml
@@ -1,0 +1,39 @@
+name: Benchmark SDK
+on:
+  push:
+    branches:
+      - master
+
+jobs:
+  benchmark:
+    name: Performance regression check
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-go@v1
+      # Run benchmark with `go test -bench` and stores the output to a file
+      - name: Run benchmark
+        run: go test -bench=. -run=^# -benchDisableLogs | tee output.txt
+      # Download previous benchmark result from cache (if exists)
+      - name: Download previous benchmark data
+        uses: actions/cache@v1
+        with:
+          path: ./cache
+          key: ${{ runner.os }}-benchmark
+      # Run `github-action-benchmark` action
+      - name: Store benchmark result
+        uses: benchmark-action/github-action-benchmark@v1
+        with:
+          # What benchmark tool the output.txt came from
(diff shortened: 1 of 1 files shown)


.github/workflows/bench.yml
```diff
@@ -0,0 +1,36 @@
+name: Benchmark
+on:
+  push:
+    branches: [ main ]
+  pull_request:
+    branches: [ main ]
```
rob-odwyer, line 6: Probably just underestimating the benchmark action, but does this not mean that the last benchmark results will be stored on PRs as well? I think we want it to store on merge to main, and compare only on PR. Does the benchmark action take care of checking the branch automatically?

.github/workflows/bench.yml
```diff
@@ -0,0 +1,36 @@
...
+        with:
+          tool: 'go'
+          output-file-path: bench_output.txt
+          external-data-json-path: ./cache/benchmark-data.json
+          summary-always: true
+          github-token: ${{ secrets.GITHUB_TOKEN }}
+          comment-on-alert: true
+          alert-threshold: 125%
+          fail-on-alert: true
+          fail-threshold: 150%
```
rob-odwyer, line 35: Is there a way to configure this action to always show the results? The warning looks good for preventing regressions, but it would be nice to show improvements as well.

If I understand the docs correctly, those above 3 settings are just about configuring failures, right?

rob-odwyer (approved):

.github/workflows/bench.yml
chris-hoefgen: Yeah, i was a bit confused on this TBH, I didn't have the `pull_request` config initially and effectively the action would never fire at all on a PR, but we kind of want to see the impact of a PR has on the benchmark before we merge right?

.github/workflows/bench.yml
chris-hoefgen: I spent an afternoon for a way to do this and couldn't come up with something inside of the `github-action-benchmark`, it looks like that reporting the benchmarks on the action is a "future feature"

Do you know of a generic display action we could maybe pipe the output to and show the results in a card or something?

PR #151, opened 2023-04-10
DVC-6908 - Variable for User

JamieSinn:


native_bucketing/bucketing_test.go
```diff
@@ -431,4 +431,28 @@ func TestClientData(t *testing.T) {
...
+	}
+
+	config, err := NewConfig(test_config, "")
+	require.NoError(t, err)
+
+	userVariable, err := generateBucketedVariableForUser(config, user, "json-var", nil)
+	require.NoError(t, err)
+	require.Equal(t, "615357cf7e9ebdca58446ed0", userVariable.Variation.Id)
+	require.Equal(t, "{\"hello\":\"world\",\"num\":610,\"bool\":true}", userVariable.Variable.Value)
+
```
rob-odwyer, line 457: We might want to test the case where the user doesn't match, and a few error cases too.

JamieSinn: Integrated those parts into the bucketed config test above

PR #152, opened 2023-04-10
Extract wasm logic from DVCClient into a new wrapper

rob-odwyer:
This refactors the DVCClient so that the WASM and object pool related code is now contained inside a new wrapper type, WASMLocalBucketing.

### DIFF
--- a/client.go
+++ b/client.go
@@ -13,7 +13,6 @@ import (
 	"net/url"
 	"reflect"
 	"regexp"
-	"strconv"
 	"strings"
 	"time"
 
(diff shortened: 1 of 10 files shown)


rob-odwyer:
* main
  * **PR #152** <a href="https://app.graphite.dev/github/pr/DevCycleHQ/go-server-sdk/152" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/DevCycleHQ/go-server-sdk/152?utm_source=stack-comment).

client.go
```diff
@@ -29,22 +28,27 @@ var (
...
 	isInitialized                bool
 	internalOnInitializedChannel chan bool
 }
 
+type LocalBucketing interface {
+	GenerateBucketedConfigForUser(userData string) (ret BucketedUserConfig, err error)
+	SetClientCustomData(customData []byte) error
+	Variable(user DVCUser, key string, variableType string) (variable Variable, err error)
+	Close()
+}
```
rob-odwyer, line 50: I'm probably going to need to tweak this interface so the serialization of the user and custom data only happens for the WASM flow. I think that's only needed for passing that data safely to WASM?

client_native.go
rob-odwyer, line 1: Removing these in this PR to make the tests pass cleanly without the native-specific client methods being fully implemented, but I'll add them back again in the next PR.

client_wasm.go
rob-odwyer, line 1: Fun fact - this filename didn't work because suffixing with `_wasm` tells Go to only compile it when building for the "wasm" arch.

JamieSinn: suffixes are hard

JamieSinn (approved):

PR #153, opened 2023-04-11
DVC-6789 - Config Manager

JamieSinn:
This is v1 of the config manager setup - handling an exported Setter, and an internal getter. 
I've converted over the params to not require the exported generatebucketedconfig and variableforuser functions to take in the config as an argument - and instead pull from the internal RWMutex locked variable.

The challenge with RWMutex is that we need to use a pointer to the custom struct - not just the struct itself - so I had to tweak some other functions to take in a pointer to the config.


native_bucketing/datamanager_config.go
```diff
@@ -1,17 +0,0 @@
...
-func GetConfigData(token string) (err error, data ConfigBody) {
-	if !emptyConfigData.Equals(configDataMap[token]) {
-		return nil, configDataMap[token]
-	}
-	return errors.New("No config data found for token " + token), data
-}
-
-func SetConfigData(token string, data ConfigBody) {
-	configDataMap[token] = data
-}
```
rob-odwyer, line 17: Is the SDK supposed to support multiple keys and configs? Do we test for that in the test harness?

native_bucketing/model_config_body.go
```diff
@@ -23,7 +23,7 @@ type Variable struct {
 	Type string `json:"type" validate:"regexp=^(String|Boolean|Number|JSON)$"`
 	Key  string `json:"key"`
 }
-type ConfigBody struct {
```
rob-odwyer, line 26: We might end up needing to expose this one again, depending on how the local bucketing layer in the root package works.

native_bucketing/config_manager.go
```diff
@@ -0,0 +1,32 @@
+package native_bucketing
+
+import (
+	"fmt"
+	"sync"
+)
+
+var internalConfig *configBody
+var configMutex = &sync.RWMutex{}
```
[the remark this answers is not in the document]
JamieSinn: Currently there's no implementation as such with the current SDK itself - and given the fact that there's no more immutable global memory concept - re-instantiating a new client should be the proper way to handle multiple sdk keys; which in turn would instantiate all new config managers/events/etc.

native_bucketing/datamanager_config.go
JamieSinn: The SDK currently doesn't inside a given context - but WASM did; we never implemented it as such.

native_bucketing/model_config_body.go
JamieSinn: Ideally we'd just pass the raw config byte array to this and the SDK layer has no need to understand the concept of the "ConfigBody" structure.

native_bucketing/config_manager.go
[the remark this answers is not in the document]
JamieSinn: IMO - We should keep the concept of the config to be irrelevant to the parent SDK - it should be entirely contained in the local bucketing concept to allow re-using this library elsewhere if needed.

PR #154, opened 2023-04-11
Unify model types in new api package

rob-odwyer:
This PR moves most of the API model types into a separate package, `api`, where they can be referenced by both the root package and native_bucketing.

I decided against making the native_bucketing types the canonical ones
for now, since some of them are used against the event API as well.


rob-odwyer:
* main
  * **PR #152** <a href="https://app.graphite.dev/github/pr/DevCycleHQ/go-server-sdk/152" target="_blank">Graphite</a> 
    * **PR #154** <a href="https://app.graphite.dev/github/pr/DevCycleHQ/go-server-sdk/154" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/DevCycleHQ/go-server-sdk/154?utm_source=stack-comment).

api/model_bucketed_user_config.go
```diff
@@ -9,5 +9,5 @@ type BucketedUserConfig struct {
 	Variables            map[string]ReadOnlyVariable `json:"variables"`
 	KnownVariableKeys    []float64                   `json:"knownVariableKeys"`
 
-	user *DVCUser
+	User *DVCUser `json:"-"`
```
JamieSinn, line 12: Why json property-ize this?

configuration.go
```diff
@@ -153,7 +155,7 @@ func NewConfiguration(options *DVCOptions) *HTTPConfiguration {
 		ConfigCDNBasePath: configBasePath,
 		EventsAPIBasePath: eventsApiBasePath,
 		DefaultHeader:     make(map[string]string),
-		UserAgent:         "DevCycle-Server-SDK/" + VERSION + "/go",
+		UserAgent:         "DevCycle-Server-SDK/" + api.VERSION + "/go",
```
JamieSinn, line 158: should this version variable not stay in the parent SDK package?

api/model_bucketed_user_config.go
rob-odwyer: It's explicitly excluding it from the JSON output, since I upgraded it to an exported field.

It looks like we might serialize BucketedUserConfig in events, and I assume the user field was set to private for a reason.

configuration.go
rob-odwyer: Yeah... this is a little awkward.
I moved it here because `PlatformData.Default` reads it, which would create a circular dependency if it was in the root package.
I'm not happy with it being here either though, so I'll try to refactor this to avoid that.

PR #161, opened 2023-04-19
Prevent errors from being surfaced to the client in variable calls

rob-odwyer:
We don't expect variable calls to ever return errors, or log when something unexpected happens other than a type mismatch.

This error silencing is basically in line with what the Node SDK does, but at some point we might want to look through the errors and see if any of them represent conditions where we'd still want to log something.

### DIFF
--- a/client.go
+++ b/client.go
@@ -305,13 +305,14 @@ func (c *DVCClient) Variable(userdata DVCUser, key string, defaultValue interfac
 			return variable, nil
 		}
 		bucketedVariable, err := c.localBucketing.Variable(userdata, key, variableType)
+
 		sameTypeAsDefault := compareTypes(bucketedVariable.Value, convertedDefaultValue)
 		if bucketedVariable.Value != nil && sameTypeAsDefault {
 			variable.Value = bucketedVariable.Value
 			variable.IsDefaulted = false
 		} else {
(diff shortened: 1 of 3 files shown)


client.go
```diff
@@ -305,13 +305,14 @@ func (c *DVCClient) Variable(userdata DVCUser, key string, defaultValue interfac
...
 		bucketedVariable, err := c.localBucketing.Variable(userdata, key, variableType)
+
 		sameTypeAsDefault := compareTypes(bucketedVariable.Value, convertedDefaultValue)
 		if bucketedVariable.Value != nil && sameTypeAsDefault {
 			variable.Value = bucketedVariable.Value
 			variable.IsDefaulted = false
 		} else {
 			if !sameTypeAsDefault && bucketedVariable.Value != nil {
-				warnf("Type mismatch for variable %s. Expected type %s, got %s",
+				debugf("Type mismatch for variable %s. Expected type %s, got %s",
```
ajwootto, line 315: I think this should stay as a warning, it basically indicates an implementation error that should be fixed

rob-odwyer: Sounds reasonable. The log volume issue we saw yesterday - that had a different cause, right?

chris-hoefgen (approved):

PR #164, opened 2023-04-20
Fix contains vs equals in string filters

rob-odwyer:
The implementation here was mixing up string contains vs. equals, which also slowed things down in the latter case.


native_bucketing/bucketing.go
```diff
@@ -102,7 +102,7 @@ type segmentedFeatureData struct {
 func evaluateSegmentationForFeature(config *configBody, feature ConfigFeature, user DVCPopulatedUser, clientCustomData map[string]interface{}) *Target {
 	for _, target := range feature.Configuration.Targets {
 		if _evaluateOperator(target.Audience.Filters, config.Audiences, user, clientCustomData) {
-			return &target
+			return target
```
rob-odwyer, line 105: This oddly avoids an allocation on every run. I think it's because the non-pointer target has been copied on the stack, and escapes the function here by being referenced by a pointer. So the compiler allocates new memory for it.
When it's a pointer to begin with, it can just be returned directly.

chris-hoefgen (approved):

PR #172, opened 2023-04-25
Adding some panic recovery logic

chris-hoefgen:


client.go
```diff
@@ -290,6 +291,14 @@ func (c *DVCClient) Variable(userdata DVCUser, key string, defaultValue interfac
 	baseVar := BaseVariable{Key: key, Value: convertedDefaultValue, Type_: variableType}
 	variable := Variable{BaseVariable: baseVar, DefaultValue: convertedDefaultValue, IsDefaulted: true}
 
+	defer func() {
+		if r := recover(); r != nil {
+			// Return a usable default value in a panic situation
+			result = variable
+			err = fmt.Errorf("recovered from panic in Variable eval: %s", string(debug.Stack()))
```
rob-odwyer, line 298: Is there anything in the return value from`recover` we'd want to log here as well? Or does the stack trace capture that already?

rob-odwyer: I think we should use those logging helpers for this instead of fmt as well. That way the logging can be controlled in one place.

chris-hoefgen: so recover() can return either a string or an Error object wrapping a string. Looks like it is whatever gets passed into the panic("something here") call

Looking at it again i'm wondering if the full stacktrace is excessive anyway, it would be really noisy in the logs

rob-odwyer: These are the logging helpers I mentioned:
https://github.com/DevCycleHQ/go-server-sdk/blob/main/logger.go#L29-L42
They're all using a shared logger, so it can be turned off, redirected, etc.

chris-hoefgen: ok, i miss understood what you were talking about earlier.  I'll get that tweaked and remove the traces

rob-odwyer (approved):

PR #178, opened 2023-05-01
Don't build native version with CGO

JamieSinn:
Can confirm that this no longer requires cgo with `CGO_ENABLED=0`


eventqueue_test.go
```diff
@@ -67,6 +64,12 @@ func TestEventQueue_QueueEvent_100_Flush(t *testing.T) {
 			break
 		}
 	}
+
+	if NATIVE_SDK {
+		// Native SDK has a background worker that flushes the queue. It's not inline and we need to wait for it to finish.
+		time.Sleep(10 * time.Second)
```
rob-odwyer, line 70: I'm not really ok with adding long sleeps to the Go tests. Other than a few of the event tests, they currently only take ~1s to run, and adding a 10 second sleep means they can't be run automatically on save anymore.

If there's no easy way to refactor this to block on the events worker, we could probably use `require.Eventually` for this.

eventqueue_test.go
```diff
@@ -67,6 +64,12 @@ func TestEventQueue_QueueEvent_100_Flush(t *testing.T) {
...
+	}
+
 	if httpmock.GetCallCountInfo()["POST https://events.devcycle.com/v1/events/batch"] != 10 {
 		t.Fatal("Expected 10 flushes to be forced. Got ", httpmock.GetCallCountInfo()["POST https://events.devcycle.com/v1/events/batch"])
 	}
```
rob-odwyer, line 75: This test should have been failing previously if it actually tested the FlushEventQueueSize behaviour, right? It looks like you just implemented it below.

rob-odwyer (changes requested):
CGO removal looks good, just have some questions about the event changes.

We should probably add `CGO_ENABLED=0` to the Github action step that runs the native bucketing tests, so we don't accidentally undo this later.

eventqueue_test.go
JamieSinn: yeah - this was just a random "wait til" hackaround that I setup to ensure that it flushes all events. Happy to rework in to a separate PR if needed

eventqueue_test.go
JamieSinn: Correct - but it was being skipped so it was never being tested

rob-odwyer: Ah, missed the skip removal. Thanks!

PR #184, opened 2023-05-02
DVC-7189 collect event bench metrics

chris-hoefgen:
Evaluating event payloads and track metrics to Datadog


configuration.go
chris-hoefgen, line 1: I'm really not 100% sold on this solution to the problem. The issue is that when the `init()` function is called inside of `test_helpers.go` it is generating unnecessary warning output which then cause the benchmark GitHub action to  fail as the output is generated before the logger can be disabled for the benchmark.

rob-odwyer (approved):

PR #185, opened 2023-05-03
Allow bypassing the variable default type checks; PlatformData passthroughs

JamieSinn:
Allow overriding platformdata
Allow bypassing the variable default type check
Properly pass platformdata into native code


rob-odwyer (approved):
Makes sense! The bypassed check here is still done on the PHP SDK side, right?

JamieSinn:
> 

yep

PR #189, opened 2023-05-09
DVC-7156 add VariableValue method

jonathannorris:


client_test.go
```diff
@@ -266,6 +274,11 @@ func TestClient_VariableLocalProtobuf_UserWithCustomData(t *testing.T) {
 		t.Fatal("Expected variable to be equal to expected variable")
 	}
 	fmt.Println(variable)
+
+	variableValue, err := c.VariableValue(user, "test", true)
```
JamieSinn, line 278: You're not checking the value of `err` here

client_test.go
```diff
@@ -218,6 +222,11 @@ func TestClient_VariableLocalProtobuf(t *testing.T) {
...
```
JamieSinn, line 226: You're not checking the value of `err` here

client.go
```diff
@@ -276,6 +276,15 @@ func (c *Client) AllFeatures(user User) (map[string]Feature, error) {
 	return nil, c.handleError(r, rBody)
 }
 
+func (c *Client) VariableValue(userdata User, key string, defaultValue interface{}) (result interface{}, err error) {
+	variable, err := c.Variable(userdata, key, defaultValue)
+	if err != nil {
+		return Variable{}, err
```
JamieSinn, line 282: Returning an empty variable object isn't a good idea here - as it'll default to an empty object with no default value set.

JamieSinn (changes requested):

client.go
jonathannorris: that's what we do in the `Variable` method below that I copied from.

jonathannorris: I guess I should just return whatever `Variable()` returns

client.go
```diff
@@ -279,7 +279,7 @@ func (c *Client) AllFeatures(user User) (map[string]Feature, error) {
 func (c *Client) VariableValue(userdata User, key string, defaultValue interface{}) (result interface{}, err error) {
 	variable, err := c.Variable(userdata, key, defaultValue)
 	if err != nil {
-		return Variable{}, err
+		return variable.Value, err
 	}
 
 	return variable.Value, nil
```
JamieSinn: This can just be `return variable.Value, err` as we don't care about the actual values of `err`

client.go
```diff
@@ -276,13 +276,35 @@ func (c *Client) AllFeatures(user User) (map[string]Feature, error) {
...
+
+    -@return interface{}
+*/
+func (c *Client) VariableValue(userdata User, key string, defaultValue interface{}) (result interface{}, err error) {
+	variable, err := c.Variable(userdata, key, defaultValue)
+	if err != nil {
+		return variable.Value, err
+	}
+
+	return variable.Value, nil
```
JamieSinn, line 296:
```suggestion
	return variable.Value, err
```

chris-hoefgen (approved):

JamieSinn (approved):

PR #198, opened 2023-06-01
fix casting of audienceMatch filters

rob-odwyer:
This fixes a bug where audienceMatch filters were being short-circuited because of a bad type cast in the segmentation logic. The unit tests didn't catch it because a non-pointer type was also provided there, instead of the pointer type that the real config would have been parsed into.

### DIFF
--- a/bucketing/segmentation.go
+++ b/bucketing/segmentation.go
@@ -45,15 +45,9 @@ func doesUserPassFilter(filter BaseFilter, audiences map[string]NoIdAudience, us
 		if !ok {
 			return false
 		}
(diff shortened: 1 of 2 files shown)


rob-odwyer:
* main
  * **PR #198** <a href="https://app.graphite.dev/github/pr/DevCycleHQ/go-server-sdk/198" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/DevCycleHQ/go-server-sdk/198?utm_source=stack-comment).

chris-hoefgen (approved):

PR #201, opened 2023-06-02
DVC-7598 version filter test

chris-hoefgen:


rob-odwyer (approved):
Looks like a solid set of tests, can't think of anything to add!

PR #210, opened 2023-06-05
DVC 7612: Rollouts

rob-odwyer:


bucketing/bucketing.go
```diff
@@ -98,9 +98,6 @@ func getCurrentRolloutPercentage(rollout Rollout, currentDate time.Time) float64
 }
 
 func doesUserPassRollout(rollout Rollout, boundedHash float64) bool {
-	if rollout.StartDate == time.Unix(0, 0) && rollout.Type == "" && len(rollout.Stages) == 0 {
-		return true
-	}
```
rob-odwyer, line 103: These checks were just an attempt to replicate `if (!rollout) return true`, which is redundant in both codebases and doesn't make sense here since we've already validated that the rollout isn't a nil pointer.

chris-hoefgen (approved):

PR #213, opened 2023-06-06
DVC-7613: add tests for missing distributions, variations, and variables

rob-odwyer:


bucketing/testdata/fixture_test_broken_config.json
rob-odwyer, line 1: Same as `barrenConfig` in js-sdks.

chris-hoefgen (approved):

PR #215, opened 2023-06-07
DVC-6956: Refactor config structure

rob-odwyer:
This refactor should reduce the risk of bugs due to type casting in the segmentation code, plus (I think) make it a lot easier to follow.

tl;dr instead of having procedural `evaluateOperator` and `doesUserPassFilter` functions that manually check the types of the filters/operators and then navigate through the tree, every filter and operator type implements `FilterOrOperator`, meaning you can call `Evaluate` on it to check whether the whole tree under that point matches a given user and custom data.


bucketing/datamanager_platform.go
```diff
@@ -6,7 +6,8 @@ func GetClientCustomData(sdkKey string) map[string]interface{} {
 	if data, ok := clientCustomData[sdkKey]; ok {
 		return data
 	}
-	return map[string]interface{}{}
+	// Nil maps are safe to read but not write, and this avoids an allocation
+	return nil
```
rob-odwyer, line 10: I'm not really sure why this didn't get flagged as an allocation previously, or why these changes made it suddenly show up in the benchmarks.
This seems safe to do though, since we never write to the custom data map. The other option is returning a global empty map var, which seems sketchier if we ever did accidentally mutate it.

chris-hoefgen (approved):

bucketing/datamanager_platform.go
chris-hoefgen: at least returning nil we will know if it causes problems pretty quickly

bucketing/segmentation_test.go
```diff
@@ -21,31 +21,27 @@ var brooks = api.PopulatedUser{
 	},
 }
 
+func _evaluateOperator(operator FilterOrOperator, audiences map[string]NoIdAudience, user api.PopulatedUser, clientCustomData map[string]interface{}) bool {
```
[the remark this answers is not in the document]
rob-odwyer: Yeah, TBH I was way too lazy to change all the `_evaluateOperator(...)` to `(...).evaluateOperator` in this file.

PR #219, opened 2023-07-12
feat: Use release action and semver PR requirements

JamieSinn:


.github/workflows/release.yml
```diff
@@ -1,30 +1,34 @@
...
+        GITHUB_TOKEN: ${{ inputs.github-token }}
+    - name: Get next minor version
+      id: semverbump
+      uses: WyriHaximus/github-action-next-semvers@v1.2.1
+      with:
+        version: ${{ steps.latestrelease.outputs.tag_name }}
     - name: Update appVersion in chart
       run: |
-        sed -i "s/^const VERSION = \".*\"/const VERSION = \"${TAG_VERSION#v}\"/" ./version.go
+        sed -i "s/^const VERSION = \".*\"/const VERSION = \"${{steps.semverbump.outputs.minor}}\"/" ./version.go
```
rob-odwyer, line 31: What do you think about doing this sed command inside the release action, to avoid duplicating the version logic?

Seems like it could be relatively similar across SDKs, maybe just with optional input for the pattern to match and the replacement to make.

rob-odwyer (approved):

.github/workflows/release.yml
chris-hoefgen: Well in the Java SDK I need to change 6 files

.github/workflows/release.yml
```diff
@@ -33,5 +37,7 @@ jobs:
         git add ./version.go
         git commit -m "Release ${TAG_VERSION}"
         git push origin main
-        git tag --force "${TAG_VERSION}"
-        git push --force origin ${{ github.event.release.tag_name }}
+    - name: Release SDK
+      uses: DevCycleHQ/release-action@main
```
rob-odwyer, line 41: We should pin to a specific version of the release action, not main. That way we can make breaking changes.

rob-odwyer (changes requested):

.github/workflows/release.yml
rob-odwyer: I'm thinking for our own benefit so we can iterate on the release action without having to keep all the SDK repos actions in sync.
If you think it's more likely that we end up spending more time on bumping these versions constantly, then we can leave it.

PR #228, opened 2023-10-23
feat: OpenFeature provider (DVC-7165)

rob-odwyer:
Add an OpenFeature provider implementation, and an example for using the provider. Also adds a workflow for testing all examples on PRs, similar to the Python SDK.


README.md
```diff
@@ -85,6 +85,12 @@ The following options are available when you are using the SDK in Local Bucketin
 | MaxWasmWorkers               | int           | The number of WASM worker objects in the object pool to support high-concurrency. <br>*Not applicable for Native Bucketing Library.*                                                                                            | GOMAXPROCS | 
 | UseDebugWASM                 | bool           | Configures the SDK to use a debug WASM binary to generate more detailed error reporting. Use caution when enabling this setting in production environments.<br>*Not applicable for Native Bucketing Library.*                   | false      |
 
+# OpenFeature Provider
+
+The SDK includes a beta provider that allows you to use the standardized OpenFeature interface. See [example/openfeature](example/openfeature/main.go) for an example of how to use it, and the following references for more information on OpenFeature:
+
+- [The DevCycle Go OpenFeature Provider](https://docs.devcycle.com/sdk/server-side-sdks/go/go-openfeature)
```
rob-odwyer, line 92: Need to actually create this page before merging this.

openfeature_provider.go
```diff
@@ -0,0 +1,379 @@
...
+	if errors.Is(err, ErrInvalidDefaultValue) {
+		return openfeature.NewTypeMismatchResolutionError(err.Error())
+	}
+	return openfeature.NewGeneralResolutionError(err.Error())
+}
+
+func createUserFromEvaluationContext(evalCtx openfeature.FlattenedContext) (User, error) {
+	userId := ""
+	_, exists := evalCtx["userId"]
+	if exists {
```
chris-hoefgen, line 286: I think we should prioritize using the Targeting Key over the UserID in the context as it is supposed to be the standard way to identify the target in the OpenFeature api

openfeature_provider.go
```diff
@@ -0,0 +1,379 @@
...
+
+	if len(privateCustomData) > 0 {
+		user.PrivateCustomData = privateCustomData
+	}
+
+	return user, nil
+}
+
+func setCustomDataValue(customData map[string]interface{}, key string, val interface{}) {
+	if val == nil {
```
chris-hoefgen, line 352: So apparently null is a valid value in custom data for DevCycle, I ended up explicitly adding it in the Java SDK.  I'm wondering if we should triple check that with Jonathan though since I personally think it is weird to have a map with a key that points at null.

example/openfeature/main.go
```diff
@@ -0,0 +1,73 @@
...
+		"country":           "CA",
+		"appVersion":        "1.0.0",
+		"appBuild":          "1",
+		"customData":        map[string]interface{}{"custom": "data"},
+		"privateCustomData": map[string]interface{}{"private": "data"},
+		"deviceModel":       "Macbook",
+	})
+
+	// Retrieving an object variable with a default value
+	value, err := client.ObjectValue(context.Background(), "test-json-variable", map[string]interface{}{"value": "default"}, evalCtx)
```
chris-hoefgen, line 48: I like showing multiple examples with different data types. I'll have to pull that into the java & python examples as well.

testdata/fixture_small_config.json
chris-hoefgen, line 1: Might be worth adding another variable that is explicitly a float to make sure those number conversions work as expected

chris-hoefgen (changes requested):

openfeature_provider.go
rob-odwyer: Interesting - that's good to check for sure.
I also wondered whether nil was supposed to be a valid object value to return? I don't think it is currently supported given how the Client handles them (defaults if it's nil).

testdata/fixture_small_config.json
rob-odwyer: That's a good call - I'll add that.

openfeature_provider.go
```diff
@@ -0,0 +1,383 @@
...
+		return openfeature.BoolResolutionDetail{
+			Value:                    defaultValue,
+			ProviderResolutionDetail: openfeature.ProviderResolutionDetail{Reason: openfeature.DefaultReason},
+		}
+	default:
+		return openfeature.BoolResolutionDetail{
+			Value: defaultValue,
+			ProviderResolutionDetail: openfeature.ProviderResolutionDetail{
+				Reason:          openfeature.ErrorReason,
+				ResolutionError: openfeature.NewTypeMismatchResolutionError("Variable result is nil"),
```
Kavindu-Dodan, line 81: I was checking the PR from the OpenFeature point of view and came across this error message. The error message here needs improvement as the `nil` case is already handled. 

Otherwise the responses are compliant with OF :)

client.go
```diff
@@ -149,6 +149,10 @@ func NewClient(sdkKey string, options *Options) (*Client, error) {
 	return c, nil
 }
 
+func (c *Client) IsLocalBucketing() bool {
+	return !c.DevCycleOptions.EnableCloudBucketing
+}
+
```
JamieSinn, line 155: Should we expose this to the public? I guess it's just a helper function for us really?

README.md
```diff
@@ -85,6 +85,17 @@ The following options are available when you are using the SDK in Local Bucketin
 | MaxWasmWorkers               | int           | The number of WASM worker objects in the object pool to support high-concurrency. <br>*Not applicable for Native Bucketing Library.*                                                                                            | GOMAXPROCS | 
 | UseDebugWASM                 | bool           | Configures the SDK to use a debug WASM binary to generate more detailed error reporting. Use caution when enabling this setting in production environments.<br>*Not applicable for Native Bucketing Library.*                   | false      |
 
+# OpenFeature Support
+
+This SDK provides an implementation of the [OpenFeature](https://openfeature.dev/) Provider interface. Use the `getOpenFeatureProvider()` method on the DevCycle SDK client to obtain a provider for OpenFeature.
```
JamieSinn, line 90:
```suggestion
This SDK provides an implementation of the [OpenFeature](https://openfeature.dev/) Provider interface. Use the `OpenFeatureProvider()` method on the DevCycle SDK client to obtain a provider for OpenFeature.
```

client.go
JamieSinn: Nevermind - I see it later on.

openfeature_provider.go
```diff
@@ -0,0 +1,409 @@
...
+	}
+	user := User{
+		UserId: userId,
+	}
+
+	customData := make(map[string]interface{})
+	privateCustomData := make(map[string]interface{})
+
+	for key, value := range evalCtx {
+		if value == nil {
```
JamieSinn, line 337: Do we not want to allow setting a custom data property to be null without explicitly calling it out as customdata?

JamieSinn: ie -https://github.com/DevCycleHQ/dotnet-server-sdk/blob/main/DevCycle.SDK.Server.Common/Model/DevCycleUser.cs#L418C51-L418C51

rob-odwyer: Good catch on this - I found some other fun bugs in custom data handling that I'm fixing now as well.

JamieSinn: the handling is weird as heck

rob-odwyer: @JamieSinn Think I've made it cover everything now, mind taking another look?

JamieSinn (approved):
I'm liking the testing setup here, might want to copy to the others

PR #230, opened 2023-12-15
fix: the notEqual operator for numbers when comparing against more than one value

elliotCamblor:
- Fix the != comparator for numbers so that it functions properly when comparing against multiple number values
- An incidental change is that the != comparator will now always return true if the values array is empty for numbers. This shouldn't be a valid filter state anyway, but worth mentioning.


JamieSinn (approved):

elliotCamblor:
> Make sure to update the WASM binary in here too - as we still provide the WASM bucketing option

Good call, didnt realize that

PR #234, opened 2024-02-23
feat: SSE Client

JamieSinn:


JamieSinn:
https://github.com/DevCycleHQ/test-harness/pull/420 - this PR needs to be merged first.

client.go
```diff
@@ -288,6 +303,16 @@ func (c *Client) Variable(userdata User, key string, defaultValue interface{}) (
 	}()
 
 	if c.IsLocalBucketing() {
+		if !c.hasConfig() {
```
ajwootto, line 306: is this being re-added here? I removed it because it's being handled properly in the local bucketing package

JamieSinn: ah - rebase mucked this one 
- lemme fix
-

JamieSinn: Fixed

configmanager.go
```diff
@@ -50,15 +55,32 @@ func NewEnvironmentConfigManager(
...
+		return fmt.Errorf("server Sent Events URI not set. Aborting SSE connection. Falling back to polling")
+	}
+	return e.sseManager.StartSSE()
+}
+
 func (e *EnvironmentConfigManager) StartPolling(
 	interval time.Duration,
 ) {
-	e.ticker = time.NewTicker(interval)
+	if e.ticker == nil {
```
ajwootto, line 81: is this to prevent problems if the StartPolling method is called more than once? seems like this whole method should be guarded with a "already ran" boolean otherwise we'd be starting another goroutine watching the channels each time

JamieSinn: This prevents an existing race condition - we don't need to guard it with an already ran - as the ticker is basically that. See the case handling below.

configmanager.go
```diff
@@ -178,6 +200,13 @@ func (e *EnvironmentConfigManager) setConfig(config []byte, eTag, rayId, lastMod
 		return err
 	}
 
+	err = json.Unmarshal(e.GetRawConfig(), &e.minimalConfig)
+	if err != nil {
+		return err
+	}
+	if e.minimalConfig != nil && e.minimalConfig.SSE != nil {
+		e.options.AdvancedOptions.RealtimeUpdatesURI = fmt.Sprintf("%s%s", e.minimalConfig.SSE.Hostname, e.minimalConfig.SSE.Path)
```
ajwootto, line 208: does this really need to be stored separately in this struct? can't it be read directly from the stored config in localbucketing so that it doesn't ever have a chance of being different than what's in that config?

JamieSinn: Yes - because we need to override it for SDK Proxy and Test harness functionality.

client.go
```diff
@@ -142,6 +151,11 @@ func NewClient(sdkKey string, options *Options) (*Client, error) {
 			}()
 		}
 	}
+
+	c.sseManager = &SSEManager{
```
ajwootto, line 155: confused why this is here in addition to being set to a (different) copy on the configmanager

ssemanager.go
```diff
@@ -0,0 +1,111 @@
...
+	Stream        *eventsource.Stream
+	eventChannel  chan eventsource.Event
+	errorHandler  eventsource.StreamErrorHandler
+}
+
+type sseEvent struct {
+	Id        string  `json:"id"`
+	Timestamp float64 `json:"timestamp"`
+	Channel   string  `json:"channel"`
+	Data      string  `json:"data"`
```
ajwootto, line 22: can this not just be treated as a byte array to begin with rather than the casting happening below?

ssemanager.go
```diff
@@ -0,0 +1,111 @@
...
+	if err != nil {
+		return
+	}
+
+	err = json.Unmarshal([]byte(event.Data), &message)
+	return
+}
+
+func (m *SSEManager) receiveSSEMessages() {
+	//nolint:all
```
ajwootto, line 82: whats violating lint here?

JamieSinn: The `for; switch` setup - it wants you to do a `for ... range` - but that explicitly isn't the intent - as we want an infinite loop not a single loop

client.go
JamieSinn: Leftover from testing. Fixed the reference

configmanager.go
ajwootto: but it's overridden for those by changing the response of the config

ssemanager.go
```diff
@@ -0,0 +1,111 @@
...
+	for {
+		select {
+		case event, ok := <-m.Stream.Events:
+			if !ok {
+				break
```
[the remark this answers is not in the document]
JamieSinn: The receiver isn't ended - this is to handle when an invalid event is received. It just skips that message.event. 
I haven't seen this hit yet - but it's part of the recommended impl for the LD eventsource client

configmanager.go
ajwootto: not sure what you mean, if this function runs more than once you'd have multiple goroutines listening on the ticker channel wouldnt you?

ssemanager.go
ajwootto: seems like it's fine with it in other places, is it just because it's inside a "plain" function rather than being inlined inside a go func?

configmanager.go
JamieSinn: in the client functionality yes - but in a server format this is how we have the uri different than the local config URI. 

In what case do you see this being different? I don't see any reasonable case where that could happen.

ssemanager.go
JamieSinn: Probably

ssemanager.go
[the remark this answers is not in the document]
ajwootto: wouldn't `break` end the infinite loop that's monitoring the stream? wouldn't you want "continue"?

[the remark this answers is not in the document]
JamieSinn: Break exists the most inner statement - in this case, the select statement. Return would exist the loop

configmanager.go
ajwootto: but why would we need the uri to be different than what's already in the config? 

the case I see is just unintentional bugs down the line, similar to the existing one we had that was caused by trying to store the etag separately. It's just defensive coding and reducing the number of possible states

ssemanager.go
JamieSinn: The spec requires it to be a string. Byte array would cause issues regarding serialization from the eventsource. This is a clone of their property set

configuration.go
```diff
@@ -113,6 +118,15 @@ func (o *Options) CheckDefaults() {
 	} else if o.FlushEventQueueSize > 50000 {
 		o.FlushEventQueueSize = 50000
 	}
+	if o.AdvancedOptions.RealtimeUpdatesQueueSize <= 0 {
```
ajwootto, line 121: what is this / where is it used?

configuration.go
```diff
@@ -113,6 +118,15 @@ func (o *Options) CheckDefaults() {
...
+		o.AdvancedOptions.RealtimeUpdatesQueueSize = 100
+	}
+	if o.AdvancedOptions.RealtimeUpdatesTimeout <= time.Second*5 {
+		o.AdvancedOptions.RealtimeUpdatesTimeout = time.Second * 5
+	}
+	if o.AdvancedOptions.RealtimeUpdatesBackoff <= time.Minute*1 {
```
ajwootto, line 127: is this really something that we want to be user configurable? feels like it could have a negative impact on our infrastructure

configuration.go
JamieSinn: SDK Proxy uses this - though it could probably be moved - it's possibly needed pending implementation

configuration.go
JamieSinn: We can bound it - but yes - because this affects how long it takes for the polling to kick in.

ajwootto: yeah but isn't that also something that should be up to us to control?

client_test.go
```diff
@@ -128,40 +128,6 @@ func TestClient_VariableLocalNumber(t *testing.T) {
 	fmt.Println(variableValue)
 }
 
-func TestClient_VariableLocalNumberWithNilDefault(t *testing.T) {
```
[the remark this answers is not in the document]
JamieSinn: ugh - more rebase pain. This was originally deleting the protobuf stuff

configmanager.go
JamieSinn: The ticker reading would listen at the same rate - preventing a race condition - but realistically we could also just return out of the function inside of this check too.

ajwootto: yeah i just dont see why we shouldnt prevent the possibility of multiple parallel goroutines

configmanager.go
JamieSinn: hm - I would have to substantially rewrite how the proxy is intending to work by doing this. The main issue is that the config itself is not exposed (in a readable/non-byte array format) to the SDK Proxy. This property is.

client.go
```diff
@@ -121,7 +122,15 @@ func NewClient(sdkKey string, options *Options) (*Client, error) {
 		}
 
 		c.configManager = NewEnvironmentConfigManager(sdkKey, c.localBucketing, options, c.cfg)
-		c.configManager.StartPolling(options.ConfigPollingIntervalMS)
+		if c.DevCycleOptions.DisableRealtimeUpdates {
+			c.configManager.StartPolling(options.ConfigPollingIntervalMS)
+		} else {
+			err = c.configManager.StartSSE()
```
ajwootto, line 128: so isn't this going to block client construction until an SSE connection is established? whereas before the initialization would have been "synchronous" with the polling started in the background, returning to the caller in an "uninitalized" state immediately?

wouldn't it be better for the establishment of the connection here to also happen in the background?

configmanager.go
ajwootto: not sure what you mean by that. Can't this just provide a getter to the underlying local bucketing config?

client.go
JamieSinn: pretty much yes - because a failed connection is delaying the same way a failed config pull would be delayed.

This is again, v1, and we want to test it out with a few failure cases to create test harness setups around it to understand what is the ideal workflow.

I don't want to release it as such - but i do want to get it into main so we can do easier/more real testing on it before we finalize the functionality and use this as the reference impl.

configmanager.go
JamieSinn: I'd prefer not to. This is more an external implementation for the SDK proxy functionality than anything.

configmanager.go
JamieSinn: I'll skip out if the ticker is running already

client.go
ajwootto: failed config pulls dont delay anything if `OnInitializedChannel` is passed, construction becomes non-blocking. This actually breaks the `OnInitializedChannel` option as well since it forces an "initial fetch" to happen here

client.go
```diff
@@ -121,7 +122,17 @@ func NewClient(sdkKey string, options *Options) (*Client, error) {
...
+			if err != nil {
+				util.Warnf("Error initializing SSE, defaulting to polling: %v", err)
+				c.configManager.StartPolling(options.ConfigPollingIntervalMS)
+			}
+		}
+
+		c.sseManager = c.configManager.sseManager
 
 		if c.DevCycleOptions.OnInitializedChannel != nil {
```
ajwootto, line 137: this stuff seems to not be doing what its expected to do now, since `initialFetch` has already happened

configmanager.go
ajwootto: but what is this supposed to represent? You're mutating the options object each time a new config is received, which already seems kinda sketchy because options should be immutable after initialization. But then it's like mutating a shared options pointer here which is also being used in the SSE manager (I think)? But that isn't directly clear since there isn't one "owner" of these options, so it's much harder to trace the side effects of mutation here

configuration.go
ajwootto: if its not meant for end users maybe it should go in an "InternalOptions"?

ssemanager.go
```diff
@@ -0,0 +1,111 @@
...
+	LastModified float64 `json:"lastModified,omitempty"`
+	Type_        string  `json:"type,omitempty"`
+}
+
+func (m *sseMessage) LastModifiedDuration() time.Duration {
+	return time.Duration(m.LastModified) * time.Millisecond
+}
+
+func newSSEManager(configManager *EnvironmentConfigManager, options *Options) *SSEManager {
+	if options == nil {
```
ajwootto, line 36: if options is nil then this is constructed with a different reference than the one configmanager is using, which breaks the whole "mutate options to give this its connection URL" thing. Another reason I think it should come from the config

client.go
JamieSinn: That's one of the things i want to narrow down how we want to handle after - but I need to test in a non-custom branch to better integrate. Specifically - how do we provide an "onupdated" equivalent channel too - and whether that should replace this initialized channel.

client.go
JamieSinn: Yes - see below

configmanager.go
JamieSinn: Another piece of functionality to think about later - mid-process SSE failure/keepalive checks failing - this will need to have a method to re-start polling after initialization

configmanager.go
JamieSinn: I'll see what I can do to rework this later while I keep building the rest of the infra surrounding it.

configuration.go
JamieSinn: Gonna change the implementation on this - and moving it around

api/model_client.go
```diff
@@ -0,0 +1,20 @@
+package api
+
+type ClientEvent struct {
```
jonathannorris, line 3: Why is the file named `model_client.go`? seems like this is all events stuff?

api/model_config.go
```diff
@@ -0,0 +1,10 @@
+package api
+
+type MinimalConfig struct {
```
jonathannorris, line 3: also the file name `mode_config` here feels a bit generic. This all seems more SSE related.

api/model_client.go
JamieSinn: this was an old name from when I had it architected differently - it was supposed to be called `model_clientevents` - I was thinking about changing it to something like `model_internal` or something?

api/model_config.go
JamieSinn: yeah - easy changed

api/model_event.go
```diff
@@ -17,6 +9,7 @@ const (
 	EventType_AggVariableEvaluated = "aggVariableEvaluated"
 	EventType_VariableDefaulted    = "variableDefaulted"
 	EventType_AggVariableDefaulted = "aggVariableDefaulted"
+	EventType_ConfigUpdated        = "sdkConfig"
```
jonathannorris, line 12: It's not specifically for config updates, more for config responses. but if this is just an internal name its fine.

JamieSinn: Yeah this is internal - I had `ConfigUpdated` as in the sense that the config has been updated - as that's when the event is being triggered

JamieSinn: It can also be easily renamed to `SDKConfig` without an issue - it's entirely internal

configmanager_test.go
```diff
@@ -66,25 +70,56 @@ func TestEnvironmentConfigManager_fetchConfig_success(t *testing.T) {
...
+	}
+	if manager.GetETag() != "TESTING" {
+		t.Fatal("cm.configEtag != TESTING")
+	}
+	if manager.sseManager == nil {
+		t.Fatal("cm.sseManager == nil")
+	}
+	require.Eventually(t, func() bool {
+		return manager.SSEConnected.Load()
+	}, 3*time.Second, 10*time.Millisecond)
```
JamieSinn, line 105: Make sure that the connection is made to SSE

configmanager_test.go
```diff
@@ -66,25 +70,56 @@ func TestEnvironmentConfigManager_fetchConfig_success(t *testing.T) {
...
-		errorResponseChain(error500Response, CONFIG_RETRIES),
+	httpmock.RegisterResponder("GET", "https://config-cdn.devcycle.com/config/v1/server/"+sdkKey+".json",
+		errorResponseChain(sdkKey, error500Response, CONFIG_RETRIES),
 	)
 
 	localBucketing := &recordingConfigReceiver{}
-	manager := NewEnvironmentConfigManager(test_environmentKey, localBucketing, test_options, NewConfiguration(test_options))
-
+	manager, _ := NewEnvironmentConfigManager(sdkKey, localBucketing, nil, test_options, NewConfiguration(test_options))
+	defer manager.Close()
```
JamieSinn, line 120: we never cleaned up our instances - and it caused some delays/cleanup issues on test spindown

configuration.go
```diff
@@ -14,38 +14,6 @@ import (
...
-	UserName string `json:"userName,omitempty"`
-	Password string `json:"password,omitempty"`
-}
-
-// APIKey provides API key based authentication to a request passed via context using ContextAPIKey
-type APIKey struct {
-	Key    string
-	Prefix string
-}
-
```
JamieSinn, line 48: This was leftover generated from openapi - never used.

configuration.go
```diff
@@ -58,11 +26,12 @@ type Options struct {
 	RequestTimeout               time.Duration `json:"requestTimeout,omitempty"`
 	DisableAutomaticEventLogging bool          `json:"disableAutomaticEventLogging,omitempty"`
 	DisableCustomEventLogging    bool          `json:"disableCustomEventLogging,omitempty"`
+	EnableBetaRealtimeUpdates    bool          `json:"enableRealtimeUpdates,omitempty"`
 	MaxEventQueueSize            int           `json:"maxEventsPerFlush,omitempty"`
 	FlushEventQueueSize          int           `json:"minEventsPerFlush,omitempty"`
 	ConfigCDNURI                 string
 	EventsAPIURI                 string
-	OnInitializedChannel         chan bool
+	ClientEventHandler           chan api.ClientEvent
```
JamieSinn, line 34: Renamed because this passes more than just init events now

event_manager.go
```diff
@@ -165,7 +192,7 @@ func (e *EventManager) flushEventPayload(
 	req.Header.Set("Content-Type", "application/json")
 	req.Header.Set("Accept", "application/json")
 
-	resp, err = e.cfg.HTTPClient.Do(req)
+	resp, err = e.httpClient.Do(req)
```
JamieSinn, line 195: we previously didn't have a direct ref to the http client - this fixed that

example/local/main.go
```diff
@@ -21,13 +21,9 @@ func main() {
...
-		EnableCloudBucketing:         false,
-		EventFlushIntervalMS:         0,
-		ConfigPollingIntervalMS:      10 * time.Second,
-		RequestTimeout:               10 * time.Second,
-		DisableAutomaticEventLogging: false,
-		DisableCustomEventLogging:    false,
+		EventFlushIntervalMS:    0,
+		ConfigPollingIntervalMS: 10 * time.Second,
+		RequestTimeout:          10 * time.Second,
 	}
```
JamieSinn, line 27: Just cleaning up defaults

example/openfeature/main.go
```diff
@@ -52,18 +52,16 @@ func main() {
...
-		log.Printf("%v = true, feature is enabled", booleanVariable)
-	} else {
-		log.Printf("%v = false, feature is disabled", booleanVariable)
+	} else if exampleValue != "DEFAULT" {
+		log.Printf("%v = %s, feature is enabled", stringVariable, exampleValue)
 	}
 
 	// Retrieving a string variable along with the resolution details
-	details, err := client.StringValueDetails(context.Background(), "doesnt-exist", "default", evalCtx)
+	details, err := client.StringValueDetails(context.Background(), "doesnt-exist", "DEFAULT", evalCtx)
```
JamieSinn, line 64: The other example tests are using the same variable - and it's actually a string - caused this test to fail when I actually moved it to a stable project - not rob's personal project.

openfeature_provider_test.go
```diff
@@ -3,33 +3,26 @@ package devcycle
...
-	provider := getProviderForConfig(t, test_config, false)
+	provider := getProviderForConfig(t, false)
 	require.Equal(t, "DevCycleProvider Local", provider.Metadata().Name)
-	provider = getProviderForConfig(t, test_config, true)
+	provider = getProviderForConfig(t, true)
 	require.Equal(t, "DevCycleProvider Cloud", provider.Metadata().Name)
 }
 
-func Test_createUserFromEvaluationContext_NoUserID(t *testing.T) {
+func TestOFcreateUserFromEvaluationContext_NoUserID(t *testing.T) {
```
JamieSinn, line 20: Renamed this to see a difference between it in debugging and test overviews

openfeature_provider_test.go
```diff
@@ -136,23 +129,34 @@ func Test_setCustomDataValue(t *testing.T) {
...
 
-func getProviderForConfig(t *testing.T, config string, cloudBucketing bool) openfeature.FeatureProvider {
+func getProviderForConfig(t *testing.T, cloudBucketing bool) openfeature.FeatureProvider {
 	t.Helper()
+	sdkKey, _ := httpConfigMock(200)
+
+	client, err := NewClient(sdkKey, &Options{
+		EnableCloudBucketing: cloudBucketing,
+	})
+	require.NoError(t, err)
```
JamieSinn, line 139: more mock setup for openfeature - same as regular tests

ssemanager.go
```diff
@@ -0,0 +1,169 @@
...
+	defer func() {
+		m.configManager.InternalClientEvents <- sseClientEvent
+	}()
+	sse, err := eventsource.SubscribeWithURL(url,
+		eventsource.StreamOptionReadTimeout(m.options.RequestTimeout),
+		eventsource.StreamOptionCanRetryFirstConnection(m.options.RequestTimeout),
+		eventsource.StreamOptionErrorHandler(m.errorHandler),
+		eventsource.StreamOptionUseBackoff(m.options.RequestTimeout),
+		eventsource.StreamOptionUseJitter(0.25),
+		eventsource.StreamOptionHTTPClient(m.cfg.HTTPClient))
```
JamieSinn, line 83: I'm not sure if I want to keep this having no clone direct property for http config/client - but having its own client might not be useful/helpful with regard to custom deployments and proxy usage.

ssemanager.go
```diff
@@ -0,0 +1,169 @@
...
+
+	err = json.Unmarshal([]byte(event.Data), &message)
+	return
+}
+
+func (m *SSEManager) receiveSSEMessages() {
+	for {
+		if m.stream == nil || m.context.Err() != nil {
+			return
+		}
```
JamieSinn, line 116: This needs to check each loop that the stream isn't nuked by a shutdown, or never connected as an error state

ssemanager.go
```diff
@@ -0,0 +1,169 @@
...
+						Status:    "",
+						Error:     nil,
+					}
+				}
+			}
+			return nil
+		}()
+		if err != nil {
+			return
+		}
```
JamieSinn, line 155: If we receive a shutdown event - we need to be able to return with an error and abort out of the infinite loop - so we need to wrap the return check in a simple error check - it could really just be as simple as a boolean - but this allows for better tracking/debugging in the future

ssemanager_test.go
```diff
@@ -0,0 +1,26 @@
+package devcycle
+
+import (
+	"testing"
+)
+
+const (
+	test_sseFullData = "{\"id\":\"yATzPE/mOzgY:0\",\"timestamp\":1712853334259,\"channel\":\"dvc_server_4fedfbd7a1aef0848768c8fad8f4536ca57e0ba0_v1\",\"data\":\"{\\\"etag\\\":\\\"\\\\\\\"714bc6a9acb038971923289ee6ce665b\\\\\\\"\\\",\\\"lastModified\\\":1712853333000}\",\"name\":\"change\"}"
+)
```
JamieSinn, line 9: This is a full SSE message.  - escaped for parsing

testdata/fixture_small_config_sse.json
```diff
@@ -0,0 +1,162 @@
...
+  "variableHashes": {
+    "test": 2447239932,
+    "test-number-variable": 3332991395,
+    "test-string-variable": 957171234,
+    "test-json-variable": 2814889459
+  },
+  "sse": {
+    "hostname": "https://sse.devcycle.com",
+    "path": "/v1/sse"
+  }
```
JamieSinn, line 161: This just adds the SSE property to the config - which I guess we could just add to the regular configs instead?

ssemanager.go
```diff
@@ -0,0 +1,169 @@
...
+		},
+		cfg: cfg,
+	}
+	sseManager.context, sseManager.stopEventHandler = context.WithCancel(context.Background())
+
+	return sseManager, nil
+}
+
+func (m *SSEManager) connectSSE(url string) (err error) {
+	if m.stream != nil {
```
ajwootto, line 64: add a comment about closing previous connection and why we have to do that

configmanager.go
```diff
@@ -22,66 +24,186 @@ type ConfigReceiver interface {
...
+					e.pollingManager.stopPolling()
+				}
+				e.SSEConnected.Store(true)
+
+			case api.ClientEventType_ConfigUpdated:
+				eventData := event.EventData.(map[string]string)
+
+				if url, ok := eventData["sseUrl"]; ok && e.options.EnableBetaRealtimeUpdates && e.sseManager != nil {
+					// Reconnect SSE
+					if e.sseManager.url != url {
```
ajwootto, line 143: need to handle more reconnection cases and continue having backup polling
-when SSE is not currently connected (never successfully connected or was interrupted) poll at a "fast enough" rate like 1 minute
-when SSE is connected, continue background polling at 10 minute intervals or so

this makes sure we dont miss config updates occurring during periods of SSE being interrupted, or when a message is just missed for whatever reason

configmanager.go
```diff
@@ -22,66 +24,186 @@ type ConfigReceiver interface {
...
+						Error:     err,
+					}
+				}
 
-	return configManager
+			case api.ClientEventType_InternalSSEConnected:
+				if e.pollingManager != nil {
+					e.pollingManager.stopPolling()
+				}
+				e.SSEConnected.Store(true)
```
ajwootto, line 136: track this state inside SSE manager instead

configmanager.go
```diff
@@ -22,66 +24,186 @@ type ConfigReceiver interface {
...
+					if currentLastModified.After(minimumLastUpdated) {
+						// Skip fetching config if the current config is newer than the minimumLastUpdated
+						continue
+					}
+				}
+
+				err := e.fetchConfig(CONFIG_RETRIES, minimumLastUpdated)
+				if err != nil {
+					util.Warnf("Error fetching config: %s\n", err)
+					e.InternalClientEvents <- api.ClientEvent{
```
ajwootto, line 110: this means we just give up on fetching the config we were notified about? this is also solved with the backup polling mechanism

client.go
```diff
@@ -120,28 +121,32 @@ func NewClient(sdkKey string, options *Options) (*Client, error) {
...
 		}
-	} else {
-		util.Infof("Using Cloud Bucketing")
-		if c.DevCycleOptions.OnInitializedChannel != nil {
-			go func() {
-				c.DevCycleOptions.OnInitializedChannel <- true
-			}()
+
+		// If SSE is enabled - the first config pull will trigger SSE to be started.
+		if !c.DevCycleOptions.EnableBetaRealtimeUpdates {
```
ajwootto, line 143: move this to be inside configManager, all the logic dealing with what "mode" its in and switching between them should be in one place so its easier to understand

configmanager.go
```diff
@@ -112,13 +236,31 @@ func (e *EnvironmentConfigManager) fetchConfig(numRetriesRemaining int) (err err
 		return err
 	}
 	defer resp.Body.Close()
+
+	if len(minimumLastModified) > 0 {
+		respLastModified := resp.Header.Get("Last-Modified")
+		if respLastModified == "" {
+			return e.fetchConfig(numRetriesRemaining, minimumLastModified...)
```
ajwootto, line 243: this could lead to endless recursion if the CDN never responds with an appropriate config

api/model_event.go
```diff
@@ -17,9 +9,29 @@ const (
...
+const (
+	ClientEventType_Initialized                ClientEventType = "initialized"
+	ClientEventType_Error                      ClientEventType = "error"
+	ClientEventType_ConfigUpdated              ClientEventType = "configUpdated"
+	ClientEventType_RealtimeUpdates            ClientEventType = "realtimeUpdates"
+	ClientEventType_InternalSSEFailure         ClientEventType = "internalSSEFailure"
+	ClientEventType_InternalNewConfigAvailable ClientEventType = "internalNewConfigAvailable"
+	ClientEventType_InternalSSEConnected       ClientEventType = "internalSSEConnected"
+)
+
```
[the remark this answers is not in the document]
ajwootto: maybe add that context to the code as a comment

ssemanager.go
```diff
@@ -0,0 +1,188 @@
...
+				Status:    "failure",
+				Error:     m.context.Err(),
+			}
+			return
+		}
+		err := func() error {
+			select {
+			case <-m.context.Done():
+				m.Connected.Store(false)
+				return fmt.Errorf("SSE - Stopping SSE polling")
```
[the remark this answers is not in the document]
JamieSinn: Technically in a few cases it's not, but canceling the context is almost never done without a reason.

ssemanager.go
```diff
@@ -0,0 +1,188 @@
...
+			case event, ok := <-m.eventChannel:
+				if !ok {
+					return nil
```
ajwootto, line 137: what is this error case?

JamieSinn: If the event channel is closed

configmanager.go
```diff
@@ -81,10 +195,11 @@ func (e *EnvironmentConfigManager) StartPolling(
 }
 
 func (e *EnvironmentConfigManager) initialFetch() error {
+
 	return e.fetchConfig(CONFIG_RETRIES)
 }
 
-func (e *EnvironmentConfigManager) fetchConfig(numRetriesRemaining int) (err error) {
+func (e *EnvironmentConfigManager) fetchConfig(numRetriesRemaining int, minimumLastModified ...time.Time) (err error) {
```
ajwootto, line 202: is this a way of doing optional params?

JamieSinn: Yeah - it creates an "Array" of length 0 if there are no params passed in

configmanager.go
```diff
@@ -115,14 +234,15 @@ func (e *EnvironmentConfigManager) fetchConfig(numRetriesRemaining int) (err err
 		return err
 	}
 	defer resp.Body.Close()
```
ajwootto, line 236: should this be checking the last modified date of the response like the node SDK is?

JamieSinn: No - that's handled through the setconfig

JamieSinn: Oh - that retry logic you told me would create an infinite retry - so it was removed?

ajwootto: it would be infinite if there wasn't a limit to the number of retries

We just talked about it yesterday for Node, so how can we have it there and not here?

JamieSinn: re-added - it was basically a one liner

ssemanager.go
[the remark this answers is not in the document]
JamieSinn: Also - the reason is that the escape condition here is checking for an error - the error object itself isn't shown to anyone - it's internal

configmanager.go
```diff
@@ -99,30 +214,40 @@ func (e *EnvironmentConfigManager) fetchConfig(numRetriesRemaining int) (err err
...
+
 	if err != nil {
 		if numRetriesRemaining > 0 {
 			util.Warnf("Retrying config fetch %d more times. Error: %s", numRetriesRemaining, err)
 			return e.fetchConfig(numRetriesRemaining - 1)
 		}
 		return err
 	}
+
+	if lmHeader, parseError := time.Parse(time.RFC1123, resp.Header.Get("Last-Modified")); parseError == nil && len(minimumLastModified) > 0 && lmHeader.Before(minimumLastModified[0]) {
```
ajwootto, line 237: this is pretty hard to read can it be split into multiple lines

JamieSinn: Sure

configmanager.go
```diff
@@ -99,30 +214,40 @@ func (e *EnvironmentConfigManager) fetchConfig(numRetriesRemaining int) (err err
...
+		return e.fetchConfig(numRetriesRemaining-1, minimumLastModified[0])
```
ajwootto, line 238: should there be some kind of delay or backoff here to give Cloudflare time to propagate the new copy

configmanager.go
```diff
@@ -99,30 +214,45 @@ func (e *EnvironmentConfigManager) fetchConfig(numRetriesRemaining int) (err err
...
 			return e.fetchConfig(numRetriesRemaining - 1)
 		}
 		return err
 	}
+	lastModifiedHeader := resp.Header.Get("Last-Modified")
+	if lastModifiedHeader != "" {
+		lastModifiedHeaderTS, parseError := time.Parse(time.RFC1123, lastModifiedHeader)
+		if parseError == nil {
+			if len(minimumLastModified) > 0 && lastModifiedHeaderTS.Before(minimumLastModified[0]) {
+				return e.fetchConfig(numRetriesRemaining-1, minimumLastModified...)
```
ajwootto, line 241: this needs to check the number of retries remaining

JamieSinn: ah good catch

JamieSinn: fixed

ajwootto (approved):

PR #238, opened 2024-04-18
feat: add event fields

suthar26:
- add event fields for the config and client identifiers

### DIFF
--- a/bucketing/event_queue.go
+++ b/bucketing/event_queue.go
@@ -53,7 +53,7 @@ func (u *UserEventQueue) BuildBatchRecords() []api.UserEventsBatchRecord {
 	return records
 }
 
-func (agg *AggregateEventQueue) BuildBatchRecords(platformData *api.PlatformData) api.UserEventsBatchRecord {
+func (agg *AggregateEventQueue) BuildBatchRecords(platformData *api.PlatformData, clientUUID string, configEtag string) api.UserEventsBatchRecord {
 	var aggregateEvents []api.Event
 	userId, err := os.Hostname()
 	if err != nil {
@@ -81,6 +81,11 @@ func (agg *AggregateEventQueue) BuildBatchRecords(platformData *api.PlatformData
 							"_variation": variation,
 							"_feature":   feature,
 						}
+					}
+					
+					metaData["clientUUID"] = clientUUID
+					if configEtag != "" {
+						metaData["configEtag"] = configEtag
 					}
 
 					event := api.Event{
@@ -240,13 +245,13 @@ func (eq *EventQueue) QueueVariableDefaultedEvent(variableKey, defaultReason str
 	return eq.queueAggregateEventInternal(variableKey, "", "", api.EventType_AggVariableDefaulted, defaultReason)
 }
 
-func (eq *EventQueue) FlushEventQueue() (map[string]api.FlushPayload, error) {
+func (eq *EventQueue) FlushEventQueue(clientUUID string, configEtag string) (map[string]api.FlushPayload, error) {
 	eq.stateMutex.Lock()
 	defer eq.stateMutex.Unlock()
 
 	var records []api.UserEventsBatchRecord
 
-	records = append(records, eq.aggEventQueue.BuildBatchRecords(eq.platformData))
+	records = append(records, eq.aggEventQueue.BuildBatchRecords(eq.platformData, clientUUID, configEtag))
 	records = append(records, eq.userEventQueue.BuildBatchRecords()...)
 	eq.aggEventQueue = make(AggregateEventQueue)
 	eq.userEventQueue = make(UserEventQueue)
--- a/client.go
+++ b/client.go
@@ -62,6 +62,7 @@ type LocalBucketing interface {
 	InternalEventQueue
 	GenerateBucketedConfigForUser(user User) (ret *BucketedUserConfig, err error)
 	SetClientCustomData(map[string]interface{}) error
+	GetClientUUID() string
 	Variable(user User, key string, variableType string) (variable Variable, err error)
 	Close()
 }
@@ -151,6 +152,10 @@ func (c *Client) IsLocalBucketing() bool {
 
 func (c *Client) handleInitialization() {
 	c.isInitialized = true
+
+	if(c.IsLocalBucketing()){
+		util.Infof("Client initialized with local bucketing %v", c.localBucketing.GetClientUUID())
+	}
 	if c.DevCycleOptions.OnInitializedChannel != nil {
 		go func() {
 			c.DevCycleOptions.OnInitializedChannel <- true
--- a/client_native_bucketing.go
+++ b/client_native_bucketing.go
@@ -6,6 +6,7 @@ import (
 	"time"
 
 	"github.com/devcyclehq/go-server-sdk/v2/util"
+	"github.com/google/uuid"
 
 	"github.com/devcyclehq/go-server-sdk/v2/api"
 
@@ -33,9 +34,12 @@ type NativeLocalBucketing struct {
 	configMutex  sync.RWMutex
 	platformData *api.PlatformData
 	eventQueue   *bucketing.EventQueue
+	clientUUID  string
 }
 
 func NewNativeLocalBucketing(sdkKey string, platformData *api.PlatformData, options *Options) (*NativeLocalBucketing, error) {
+	clientUUID := uuid.New().String()
+
 	eq, err := bucketing.NewEventQueue(sdkKey, options.eventQueueOptions(), platformData)
 	if err != nil {
 		return nil, err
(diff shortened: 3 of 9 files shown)


suthar26:
* **#238** <a href="https://app.graphite.dev/github/pr/DevCycleHQ/go-server-sdk/238?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

This stack of pull requests is managed by Graphite. <a href="https://stacking.dev/?utm_source=stack-comment">Learn more about stacking.</a>
<h2></h2>

 Join @suthar26 and the rest of your teammates on <a href="https://graphite.dev?utm-source=stack-comment">Graphite <b>Graphite</b></a>

client.go
```diff
@@ -17,6 +17,8 @@ import (
 	"strings"
 	"time"
 
+	"github.com/google/uuid"
```
ajwootto, line 20: is this adding a new dependency?

client.go
```diff
@@ -134,6 +137,7 @@ func NewClient(sdkKey string, options *Options) (*Client, error) {
 			c.handleInitialization()
 			return c, err
 		}
+		c.clientUUID = uuid.NewString()
```
ajwootto, line 140: this wont be set since its returning before here on the line above

example/cloud/main.go
```diff
@@ -68,6 +68,8 @@ func main() {
 	event := devcycle.Event{
 		Type_:  "customEvent",
 		Target: "somevariable.key",
+		ConfigEtag: "etag",
```
ajwootto, line 71: this isn't a valid thing to make people have to do when tracking events

client.go
suthar26: no, it already exists in go.mod

api/model_event.go
```diff
@@ -29,6 +29,8 @@ type Event struct {
 	Value       float64                `json:"value,omitempty"`
 	FeatureVars map[string]string      `json:"featureVars"`
 	MetaData    map[string]interface{} `json:"metaData,omitempty"`
+	ConfigEtag 	string                 `json:"configEtag,omitempty"`
```
ajwootto, line 32: these arent going to go anywhere since these fields dont exist in the event api. For them to end up in the events DB they need to go in metadata

example/cloud/main.go
suthar26: this should not be here actually, will remove

client.go
```diff
@@ -477,8 +481,11 @@ func (c *Client) Track(user User, event Event) (bool, error) {
 		return false, errors.New("event type is required")
 	}
 
+	event.ClientId = c.clientUUID
+
 	if c.IsLocalBucketing() {
 		if c.hasConfig() {
+			event.ConfigEtag = c.configManager.configETag
```
ajwootto, line 488: this is the property difference I was talking about, since this is stored separately from the config it doesn't _necessarily_ represent the actual etag of the config being used. Would be nice if it could come from that config

client.go
```diff
@@ -477,8 +487,23 @@ func (c *Client) Track(user User, event Event) (bool, error) {
 		return false, errors.New("event type is required")
 	}
 
+	if(event.MetaData == nil) {
```
ajwootto, line 490: we don't want this in custom tracked events we specifically want it in the aggVariableDefaulted and aggVariableEvaluated events

bucketing/event_queue.go
```diff
@@ -224,30 +233,32 @@ func (eq *EventQueue) QueueEvent(user api.User, event api.Event) error {
...
 	if eq.options.DisableAutomaticEventLogging {
 		return nil
 	}
 
-	return eq.queueAggregateEventInternal(variableKey, "", "", api.EventType_AggVariableDefaulted, defaultReason)
+	return eq.queueAggregateEventInternal(variableKey, "", "", api.EventType_AggVariableDefaulted, defaultReason, clientUUID, configEtag)
 }
 
-func (eq *EventQueue) FlushEventQueue() (map[string]api.FlushPayload, error) {
+func (eq *EventQueue) FlushEventQueue(clientUUID string, configEtag string) (map[string]api.FlushPayload, error) {
```
ajwootto, line 253: it doesnt really make sense for these to be passed in here, they should come from the metadata on the events that are already stored in the queue

client.go
```diff
@@ -151,6 +154,8 @@ func (c *Client) IsLocalBucketing() bool {
 
 func (c *Client) handleInitialization() {
 	c.isInitialized = true
+	c.clientUUID = uuid.NewString()
+	c.SetClientCustomData(map[string]interface{}{"clientUUID": c.clientUUID})
```
ajwootto, line 158: this isn't a valid place to put this, since client custom data is set by the end user

bucketing/event_queue.go
suthar26: I was trying to get this data to the event_queue and this seemed like the most efficient way but I see what you mean will see if its in the metadata somewhere

client.go
suthar26: will see if I can put it elsewhere

bucketing/event_queue.go
ajwootto: well you should look at the cycle of when events get flushed etc. and whether this data needs to be on the individual events that are pushed in here or not. The SDK is supposed to flush events any time the config is about to change, so it would make sense to mark all outgoing events at that time using the config etag before the new config is set. But you'd have to verify that the SDK is actually flushing events properly when the config changes

client.go
ajwootto: you already put it on the client directly, so how come it needs to be in custom data too?

configmanager.go
```diff
@@ -137,16 +137,29 @@ func (e *EnvironmentConfigManager) fetchConfig(numRetriesRemaining int) (err err
 
 func (e *EnvironmentConfigManager) setConfigFromResponse(response *http.Response) error {
 	config, err := io.ReadAll(response.Body)
+
+	if err != nil {
+		return err
+	}
+
+	configMap := make(map[string]interface{})
```
ajwootto, line 145: what are these changes for?

event_manager.go
```diff
@@ -102,22 +102,23 @@ func (e *EventManager) QueueEvent(user User, event Event) error {
...
 	err = e.internalQueue.QueueEvent(user, event)
 	if err != nil && errors.Is(err, ErrQueueFull) {
 		return fmt.Errorf("event queue is full, dropping event: %+v", event)
 	}
 	return err
 }
 
-func (e *EventManager) QueueVariableDefaultedEvent(variableKey string, defaultReason string) error {
-	return e.internalQueue.QueueVariableDefaulted(variableKey, defaultReason)
+func (e *EventManager) QueueVariableDefaultedEvent(variableKey string, defaultReason string, clientUUID string, configEtag string) error {
```
ajwootto, line 114: you gotta decide whether these are going on the individual event observations or are gonna be passed in when events get flushed. Right now it seems to be doing both

bucketing/bucketing.go
```diff
@@ -209,13 +209,14 @@ func GenerateBucketedConfig(sdkKey string, user api.PopulatedUser, clientCustomD
 		FeatureVariationMap:  featureVariationMap,
 		VariableVariationMap: variableVariationMap,
 		Variables:            variableMap,
+		ETag:                 config.eTag,
```
ajwootto, line 212: how come this needs to be on the BucketedUserConfig?

bucketing/datamanager_platform.go
```diff
@@ -13,3 +13,11 @@ func GetClientCustomData(sdkKey string) map[string]interface{} {
 func SetClientCustomData(sdkKey string, data map[string]interface{}) {
 	clientCustomData[sdkKey] = data
 }
+
+func GetConfigEtag(sdkKey string) string {
```
ajwootto, line 17: isn't this the same as the other method defined in config_manager

client_native_bucketing.go
```diff
@@ -18,11 +19,13 @@ const NATIVE_SDK = true
...
 
 func (c *Client) setLBClient(sdkKey string, options *Options) error {
-	localBucketing, err := NewNativeLocalBucketing(sdkKey, c.platformData, options)
+	clientUUID := uuid.New().String()
+	localBucketing, err := NewNativeLocalBucketing(sdkKey, c.platformData, options, clientUUID)
 	if err != nil {
 		return err
 	}
 	c.localBucketing = localBucketing
+	c.clientUUID = clientUUID
```
ajwootto, line 28: the outer client doesn't really need another copy of this if its already stored on the local bucketing

suthar26: using this for logging after initialization is done, or should that be done on local bucketing?

ajwootto: you can read the value from the local bucketing when doing that log

suthar26: fixed

client.go
```diff
@@ -151,6 +152,11 @@ func (c *Client) IsLocalBucketing() bool {
 
 func (c *Client) handleInitialization() {
 	c.isInitialized = true
+
+	if(c.IsLocalBucketing()){
+		util.Infof("Client initialized with local bucketing %v", c.localBucketing.GetClientUUID())
+		fmt.Println("Client initialized with local bucketing", c.localBucketing.GetClientUUID())
```
ajwootto, line 158: dont think you need this one?

suthar26: yeah was added to test locally

bucketing/config_manager.go
```diff
@@ -23,7 +23,7 @@ func GetEtag(sdkKey string) string {
 	if err != nil {
 		return ""
 	}
-	return config.etag
+	return config.eTag
```
JamieSinn, line 26: etag is the correct capitalization

ajwootto (approved):

PR #240, opened 2024-04-18
add tracking for default reason

ajwootto:


bucketing/bucketing.go
```diff
@@ -214,24 +215,25 @@ func GenerateBucketedConfig(sdkKey string, user api.PopulatedUser, clientCustomD
...
-			util.Warnf("Failed to queue variable evaluated event: %s", eventErr)
+			util.Warnf("Failed to queue variable defaulted event: %s", eventErr)
 		}
+		return "", nil, err
+	}
+
+	eventErr := eventQueue.QueueVariableEvaluatedEvent(variableKey, featureId, variationId)
+	if eventErr != nil {
+		util.Warnf("Failed to queue variable evaluated event: %s", eventErr)
 	}
```
jonathannorris, line 237: why remove `!eventQueue.options.DisableAutomaticEventLogging`?

ajwootto: the check was already being done in the event queueing methods, and having it be repeated here was confusing since it wasn't being checked for all cases of an event being logged in this method, only one of them. That looked like a bug, but it was actually just redundant

bucketing/event_queue.go
```diff
@@ -53,47 +54,34 @@ func (agg *AggregateEventQueue) BuildBatchRecords(platformData *api.PlatformData
 
 	for _type, variableAggMap := range *agg {
 		for variableKey, featureAggMap := range variableAggMap {
-			if variationAggMap, ok := featureAggMap["value"]; ok {
-				if variationValue, ok2 := variationAggMap["value"]; ok2 && variationValue > 0 {
-					value := float64(variationValue)
+			for feature, _variationAggMap := range featureAggMap {
+				for variation, count := range _variationAggMap {
```
jonathannorris, line 58: Can you describe what these changes are doing?

client.go
```diff
@@ -338,7 +338,7 @@ func (c *Client) Variable(userdata User, key string, defaultValue interface{}) (
 		if !c.hasConfig() {
 			util.Warnf("Variable called before client initialized, returning default value")
 
-			err = c.eventQueue.QueueVariableDefaultedEvent(key)
+			err = c.eventQueue.QueueVariableDefaultedEvent(key, "CONFIG_MISSING")
```
JamieSinn, line 341: This should be a constant in `util`

JamieSinn (changes requested):
You have some test failures

client.go
ajwootto: there arent any other constants defined in there, is there a better place for it? it's related to the error types defined in the bucketing package

JamieSinn: You could add it to the event queue constant area?

ajwootto: can I import it from the bucketing lib where all of this is defined?

JamieSinn: Yep

JamieSinn (approved):

PR #245, opened 2024-04-22
chore: separate cloud client implementation

ajwootto:
- separate cloud client implementation from main client

### DIFF
--- a/client.go
+++ b/client.go
@@ -1,27 +1,16 @@
 package devcycle
 
 import (
-	"bytes"
 	"context"
 	"errors"
 	"fmt"
-	"io"
-	"math"
-	"math/rand"
-	"net/http"
-	"net/url"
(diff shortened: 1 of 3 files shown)


client.go
```diff
@@ -184,108 +177,27 @@ func (c *Client) GetRawConfig() (config []byte, etag string, err error) {
...
-		}
-	}
-
-	return &proto.NullableCustomData{
-		Value:  dataMap,
-		IsNull: false,
-	}
-}
-
 /*
```
JamieSinn, line 180: lets separate the protobuf cleanup from this pr

JamieSinn (changes requested):

client.go
ajwootto: the functions here were unused tho

JamieSinn: yes - but they're confusing to review

PR #252, opened 2024-04-26
fix: Send last modified header along with etag

JamieSinn:
I really debated changing the params to be a custom object instead of having umpteen params.

### DIFF
--- a/bucketing/config_manager.go
+++ b/bucketing/config_manager.go
@@ -5,6 +5,7 @@ import (
 	"sync"
 )
 
+var internalLastModified = make(map[string]string)
 var internalConfigs = make(map[string]*configBody)
 var internalRawConfigs = make(map[string][]byte)
 var configMutex = &sync.RWMutex{}
@@ -34,6 +35,15 @@ func GetRayId(sdkKey string) string {
 	return config.rayId
 }
 
+func GetLastModified(sdkKey string) string {
+	configMutex.RLock()
+	defer configMutex.RUnlock()
+	if val, ok := internalLastModified[sdkKey]; ok {
+		return val
+	}
+	return ""
+}
+
 func GetRawConfig(sdkKey string) []byte {
 	configMutex.RLock()
 	defer configMutex.RUnlock()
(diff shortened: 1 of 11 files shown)


bucketing/config_manager.go
```diff
@@ -5,6 +5,7 @@ import (
 	"sync"
 )
 
+var internalLastModified = make(map[string]string)
```
ajwootto, line 8: why do you need this map if its stored on the config object?

bucketing/config_manager.go
```diff
@@ -43,14 +53,15 @@ func GetRawConfig(sdkKey string) []byte {
 	return nil
 }
 
-func SetConfig(rawJSON []byte, sdkKey, etag string, rayId string, eventQueue ...*EventQueue) error {
+func SetConfig(rawJSON []byte, sdkKey, etag, rayId, lastModified string, eventQueue ...*EventQueue) error {
```
jonathannorris, line 56: why is `string` removed from: `etag, rayId`?

JamieSinn: go preferred format. For multiple params of the same type - you're supposed to define it on the last one - instead of each

bucketing/config_manager.go
JamieSinn: if a config object is invalid/corrupted - we need to store it elsewhere - same as the etag

ajwootto: the etag isnt stored anywhere else

if the config is invalid or corrupted then the etag that went along with it is not being used and thus isnt stored, which should be the same as this last modified thing

JamieSinn: I'm thinking of the test version - will match that

ajwootto (approved):
should there be a test harness update for this behaviour?

JamieSinn:
> should there be a test harness update for this behaviour?

I mentioned that - I think it would be wise - but given that we don't actually know what the "correct" behaviour is - I think we need to find that out before we create a test harness test blindly

ajwootto:
> > should there be a test harness update for this behaviour?
> 
> I mentioned that - I think it would be wise - but given that we don't actually know what the "correct" behaviour is - I think we need to find that out before we create a test harness test blindly

from the test harness perspective we want to test that the SDK stores and re-uses the last-modified date on subsequent requests as the If-Modified-Since header

JamieSinn:
> from the test harness perspective we want to test that the SDK stores and re-uses the last-modified date on subsequent requests as the If-Modified-Since header

Right - but what happens if you send both? Only one? etc that logic we don't know what happens yet

PR #254, opened 2024-05-01
feat: SSE Config Consolidator Event Flow

JamieSinn:
- Change our test initialization strategy to do the following:
   - Never use the same sdk key - this causes a race condition in some cases due to spinning up and destroying config mocks.
   - Start and stop the mocks only once in the test overall, help prevent the race conditions on tests only.
   - Generate test SDK keys in a helper instead of the test function
- Create a threadsafe setup for handling SSE updates and polling/sse handoffs.

### DIFF
--- a/client.go
+++ b/client.go
@@ -86,7 +86,7 @@ func NewClient(sdkKey string, options *Options) (*Client, error) {
 		return nil, err
 	}
 	if !sdkKeyIsValid(sdkKey) {
-		return nil, fmt.Errorf("Invalid sdk key. Call NewClient with a valid sdk key.")
+		return nil, fmt.Errorf("invalid sdk key %s. Call NewClient with a valid sdk key", sdkKey)
 	}
 	options.CheckDefaults()
 	cfg := NewConfiguration(options)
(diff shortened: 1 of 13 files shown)


client.go
```diff
@@ -51,10 +51,9 @@ type Client struct {
 	eventQueue      *EventManager
 	localBucketing  LocalBucketing
 	platformData    *PlatformData
-	sseManager      *SSEManager
 	// Set to true when the client has been initialized, regardless of whether the config has loaded successfully.
-	isInitialized                bool
-	internalOnInitializedChannel chan bool
+	isInitialized              bool
+	internalClientEventChannel chan api.ClientEvent
```
JamieSinn, line 56: this is currently still only used for initialized events - but can be expanded in the future.

client.go
```diff
@@ -101,15 +100,14 @@ func NewClient(sdkKey string, options *Options) (*Client, error) {
 	} else {
 		c.platformData = GeneratePlatformData()
 	}
+	c.internalClientEventChannel = make(chan api.ClientEvent, 1)
```
JamieSinn, line 103: This channel is used for both local and cloud bucketing.

configmanager.go
```diff
@@ -23,90 +24,164 @@ type ConfigReceiver interface {
...
+		go configManager.ssePollingManager()
+	}
 	return configManager
 }
 
-func (e *EnvironmentConfigManager) StartSSE() error {
-	err := e.initialFetch()
-	if err != nil {
-		return err
+func (e *EnvironmentConfigManager) ssePollingManager() {
```
JamieSinn, line 73: This is the main event loop to handle the changes to config status, SSE, and polling

PR #256, opened 2024-05-08
chore: Update depends - and go base version

JamieSinn:


jonathannorris (approved):
Does updating the base go verison limit SDK compatibility?

PR #262, opened 2024-06-17
fix: On 304 for SSE - reconnect when not connected

JamieSinn:


configmanager.go
```diff
@@ -251,6 +251,20 @@ func (e *EnvironmentConfigManager) fetchConfig(numRetriesRemaining int, minimumL
 		resp.Request = req
 		return e.setConfigFromResponse(resp)
 	case statusCode == http.StatusNotModified:
+		if e.sseManager != nil && !e.sseManager.Connected.Load() && e.minimalConfig != nil && e.minimalConfig.SSE != nil {
+			configUpdatedEvent := api.ClientEvent{
```
jonathannorris, line 255: is this a common event? should it be in a shared function somewhere?

JamieSinn: Not really - this is specific to the event handler flow

jonathannorris: how is this triggered for the 200 responses?

JamieSinn: Via the set config event; this one is specific because it's passing an existing property back - not a new url

jonathannorris (approved):

configmanager.go
JamieSinn: https://github.com/DevCycleHQ/go-server-sdk/blob/main/configmanager.go#L318

PR #263, opened 2024-07-12
fix: Ignore older config lastmodified timestamps when fetching a config.

JamieSinn:
Also set clientdate on sdkconfig events, and regular tracked events.

### DIFF
--- a/configmanager.go
+++ b/configmanager.go
@@ -226,7 +226,7 @@ func (e *EnvironmentConfigManager) fetchConfig(numRetriesRemaining int, minimumL
 	lastModified := e.localBucketing.GetLastModified()
 	storedLM, err := time.Parse(time.RFC1123, lastModified)
 	if err != nil {
-		util.Warnf("Error parsing last modified time: %s\n", err)
+		util.Warnf("Error parsing stored last modified time: %s\n", err)
 	}
 	if len(minimumLastModified) > 0 && storedLM.Before(minimumLastModified[0]) {
 		lastModified = minimumLastModified[0].Format(time.RFC1123)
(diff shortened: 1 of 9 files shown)


configmanager.go
```diff
@@ -236,9 +240,12 @@ func (e *EnvironmentConfigManager) fetchConfig(numRetriesRemaining int, minimumL
 	}
 	lastModifiedHeader := resp.Header.Get("Last-Modified")
 	if lastModifiedHeader != "" {
-		lastModifiedHeaderTS, parseError := time.Parse(time.RFC1123, lastModifiedHeader)
+		responseLastModified, parseError := time.Parse(time.RFC1123, lastModifiedHeader)
 		if parseError == nil {
-			if len(minimumLastModified) > 0 && lastModifiedHeaderTS.Before(minimumLastModified[0]) && numRetriesRemaining > 0 {
+			if storedLM, parseErr := time.Parse(time.RFC1123, e.localBucketing.GetLastModified()); parseErr == nil && storedLM.After(responseLastModified) {
```
suthar26, line 245: this is the same as `lastModified` or the `minimumLastModified` would that not be a better comparison?

configmanager.go
```diff
@@ -215,15 +215,19 @@ func (e *EnvironmentConfigManager) fetchConfig(numRetriesRemaining int, minimumL
...
-		lastModified = minimumLastModified[0].Format(time.RFC1123)
+	storedLM, err := time.Parse(time.RFC1123, lastModified)
+	if err != nil {
+		util.Warnf("Error parsing last modified time: %s\n", err)
 	}
-	if etag != "" {
-		req.Header.Set("If-None-Match", etag)
+	if len(minimumLastModified) > 0 && storedLM.Before(minimumLastModified[0]) {
+		lastModified = minimumLastModified[0].Format(time.RFC1123)
 	}
```
jonathannorris, line 224: Can you explain what this is trying to do? where is `minimumLastModified` coming from?

configmanager.go
```diff
@@ -236,9 +240,12 @@ func (e *EnvironmentConfigManager) fetchConfig(numRetriesRemaining int, minimumL
...
 	lastModifiedHeader := resp.Header.Get("Last-Modified")
 	if lastModifiedHeader != "" {
-		lastModifiedHeaderTS, parseError := time.Parse(time.RFC1123, lastModifiedHeader)
+		responseLastModified, parseError := time.Parse(time.RFC1123, lastModifiedHeader)
 		if parseError == nil {
-			if len(minimumLastModified) > 0 && lastModifiedHeaderTS.Before(minimumLastModified[0]) && numRetriesRemaining > 0 {
+			if storedLM.After(responseLastModified) {
+				return e.fetchConfig(numRetriesRemaining - 1)
```
jonathannorris, line 246: should we add a warning here?

configmanager.go
JamieSinn: Check if MinimumLastModified is passed (length check) - if it exists; then check whether or not the stored last modified timestamp is before the passed new minimum. If it is - set that as the new header to set.

configmanager.go
```diff
@@ -236,9 +240,12 @@ func (e *EnvironmentConfigManager) fetchConfig(numRetriesRemaining int, minimumL
...
+			}
+			if len(minimumLastModified) > 0 && responseLastModified.Before(minimumLastModified[0]) && numRetriesRemaining > 0 {
 				return e.fetchConfig(numRetriesRemaining-1, minimumLastModified...)
 			}
```
jonathannorris, line 250: doesn't there need to be an else statement here that if its beyond the number of retries that it doesn't try and save the old request body?

configmanager.go
JamieSinn: minlastmodified is used by SSE to pass a minimum request timestamp that we should expect.

configmanager.go
JamieSinn: we could - but nothing really will be done from it - it's an inactionable warning

configmanager.go
jonathannorris: also need a test for this ^

JamieSinn: I'll add an abort out otherways - but that's not where we would want to abort.

JamieSinn: We do - it's the test I added

jonathannorris (approved):

PR #264, opened 2024-08-12
feat: add new filters to go bucketing logic

suthar26:
- adds new filters `startWith` and `endWith` for string comparison

### DIFF
--- a/bucketing/constants.go
+++ b/bucketing/constants.go
@@ -46,6 +46,10 @@ const (
 	ComparatorNotExist     = "!exist"
 	ComparatorContain      = "contain"
 	ComparatorNotContain   = "!contain"
+	ComparatorStartWith    = "startWith"
+	ComparatorNotStartWith = "!startWith"
+	ComparatorEndWith      = "endWith"
+	ComparatorNotEndWith   = "!endWith"
 )
 
 const (
--- a/bucketing/segmentation.go
+++ b/bucketing/segmentation.go
@@ -91,18 +91,26 @@ func _checkNumbersFilter(number float64, filter *UserFilter) bool {
 func checkStringsFilter(str string, filter *UserFilter) bool {
 	operator := filter.GetComparator()
 	values := filter.CompiledStringVals
-	if operator == "=" {
+	if operator == ComparatorEqual {
 		return str != "" && stringArrayIn(values, str)
-	} else if operator == "!=" {
+	} else if operator == ComparatorNotEqual {
 		return str != "" && !stringArrayIn(values, str)
-	} else if operator == "exist" {
+	} else if operator == ComparatorExist {
 		return str != ""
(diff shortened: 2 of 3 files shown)


bucketing/constants.go
```diff
@@ -46,6 +46,10 @@ const (
 	ComparatorNotExist     = "!exist"
 	ComparatorContain      = "contain"
 	ComparatorNotContain   = "!contain"
+	ComparatorStartWith    = "startWith"
```
JamieSinn, line 49: is it `startWith` or `startsWith`

bucketing/segmentation.go
```diff
@@ -103,6 +103,14 @@ func checkStringsFilter(str string, filter *UserFilter) bool {
 		return str != "" && stringArrayContains(values, str)
 	} else if operator == "!contain" {
 		return str == "" || !stringArrayContains(values, str)
+	} else if operator == "startWith" {
```
JamieSinn, line 106: Use the constant refs instead of the string raw itself.

bucketing/segmentation.go
```diff
@@ -126,6 +134,24 @@ func stringArrayContains(substrings []string, search string) bool {
 	return false
 }
 
+func stringArrayStartsWith(prefixes []string, search string) bool {
+	for _, prefix := range prefixes {
+		if strings.HasPrefix(search, prefix) {
```
JamieSinn, line 139: Keep in mind - that `HasPrefix` and `HasSuffix` will return true if `prefix` in this arg is emptystring. 

This is something we should ensure is documented, or prevented from happening


`fmt.Println(strings.HasPrefix("Gopher", ""))`

suthar26: we dont currently catch this for `strings.Contains` either so if we add a check for `HasSuffix` and `HasPrefix` it should be added to `Contains` as well.. do you think that is a breaking change?

bucketing/constants.go
suthar26: its `startWith` and `endWith`

bucketing/segmentation.go
JamieSinn: as long as the bucketing logic is identical across platforms - we're fine - how does JS/WASM handle it?

bucketing/segmentation.go
```diff
@@ -119,7 +127,25 @@ func stringArrayIn(arr []string, search string) bool {
 
 func stringArrayContains(substrings []string, search string) bool {
 	for _, substring := range substrings {
-		if strings.Contains(search, substring) {
+		if substring != "" && strings.Contains(search, substring) {
```
JamieSinn, line 130: You can just make this a continue if it's empty string, easier and cleaner

JamieSinn (approved):

JamieSinn (approved):

PR #266, opened 2024-08-16
feat: Add support for alternate key bucketing

kaushalkapasi:
- feat: add support for alternate key bucketing rollouts and random distributions
  - supports String, Number and Boolean types for alternate keys
- chore: updates test data and test cases to ensure consistent bucketing for users with and without the alternate key defined

### DIFF
--- a/bucketing/bucketing.go
+++ b/bucketing/bucketing.go
@@ -29,7 +29,7 @@ type boundedHash struct {
 	BucketingHash float64 `json:"bucketingHash"`
 }
 
-func generateBoundedHashes(bucketingKeyValue string, targetId string) boundedHash {
+func generateBoundedHashes(bucketingKeyValue, targetId string) boundedHash {
 	var targetHash = murmurhashV3(targetId, baseSeed)
 	var bhash = boundedHash{
 		RolloutHash:   generateBoundedHash(bucketingKeyValue+"_rollout", targetHash),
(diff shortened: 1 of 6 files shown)


bucketing/bucketing.go
```diff
@@ -27,11 +29,11 @@ type boundedHash struct {
 	BucketingHash float64 `json:"bucketingHash"`
 }
 
-func generateBoundedHashes(userId, targetId string) boundedHash {
+func generateBoundedHashes(bucketingKeyValue string, targetId string) boundedHash {
```
JamieSinn, line 32:
```suggestion
func generateBoundedHashes(bucketingKeyValue, targetId string) boundedHash {
```

Multiple params of the same type only require you to define the type on the last param

bucketing/bucketing.go
```diff
@@ -41,6 +43,31 @@ func generateBoundedHash(input string, hashSeed uint32) float64 {
 	return float64(mh) / float64(maxHashValue)
 }
 
+func determineUserBucketingValue(userId string, mergedCustomData map[string]interface{}, targetBucketingKey string) string {
```
JamieSinn, line 46: you could make this a function on the user object - given it's taking params from the user object anyway

kaushalkapasi: logically, this is related to the Target and which User attribute it will use to perform a rollout / random distribution, that's why I added it here

JamieSinn: right - but just from a performance aspect you're creating two new allocations (grabbing the merged custom data each time you call this) instead of just referring to the native object prop

bucketing/testdata/fixture_test_config.json
JamieSinn, line 1: I would create a new file here to specifically call out it's a v2 config

kaushalkapasi: sounds good, will update to reflect the new field on targets

JamieSinn:
Do you want to also vbump the path for the configs with this change too?

JamieSinn (approved):

elliotCamblor:
> Do you want to also vbump the path for the configs with this change too?

@JamieSinn I think we want to vbump the config path as its own PR because thats the true breaking change, and the versioned configs feature flag isnt on for all users yet

JamieSinn:
Kk

JamieSinn (approved):

PR #267, opened 2024-08-19
feat: update config url to support v2 configs

suthar26:
- update config url and tests to v2

### DIFF
--- a/.github/workflows/run-test-harness.yml
+++ b/.github/workflows/run-test-harness.yml
@@ -12,7 +12,7 @@ jobs:
     steps:
       - uses: DevCycleHQ/test-harness@main
         env:
-          SDK_CAPABILITIES: '["cloud","edgeDB","clientCustomData","multithreading","defaultReason","etagReporting","lastModifiedHeader","sdkConfigEvent","clientUUID"]'
+          SDK_CAPABILITIES: '["cloud","edgeDB","clientCustomData","multithreading","defaultReason","etagReporting","lastModifiedHeader","sdkConfigEvent","clientUUID","v2Config"]'
         with:
           sdks-to-test: '["go"]'
           sdk-github-sha: ${{github.event.pull_request.head.sha}}
--- a/configmanager.go
+++ b/configmanager.go
@@ -4,11 +4,12 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
-	"github.com/devcyclehq/go-server-sdk/v2/api"
 	"io"
 	"net/http"
 	"sync"
 	"time"
+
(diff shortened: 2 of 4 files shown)


configmanager_test.go
```diff
@@ -113,7 +114,7 @@ func TestEnvironmentConfigManager_fetchConfig_refuseOld(t *testing.T) {
 	secondResponse := httpCustomConfigMock(sdkKey, 200, test_config, true, olderHeaders)
 	thirdResponse := httpCustomConfigMock(sdkKey, 200, test_config, true, newestHeaders)
 
-	httpmock.RegisterResponder("GET", "https://config-cdn.devcycle.com/config/v1/server/"+sdkKey+".json",
+	httpmock.RegisterResponder("GET", "https://config-cdn.devcycle.com/config/v2/server/"+sdkKey+".json",
```
JamieSinn, line 117: I'd just setup a test specific variable for the base config slug `https://config-cdn.devcycle.com/config/v2/server/` and then have that `fmt.Sprintf`'d into all the urls - instead of the janky concat we're doing now

configmanager.go
```diff
@@ -392,7 +393,7 @@ func (e *EnvironmentConfigManager) setConfig(config []byte, eTag, rayId, lastMod
 func (e *EnvironmentConfigManager) getConfigURL() string {
 	configBasePath := e.cfg.ConfigCDNBasePath
 
-	return fmt.Sprintf("%s/config/v1/server/%s.json", configBasePath, e.sdkKey)
+	return fmt.Sprintf("%s/config/v2/server/%s.json", configBasePath, e.sdkKey)
```
JamieSinn, line 396: You'll also need to do a test harness update for this

.github/workflows/run-test-harness.yml
```diff
@@ -12,7 +12,7 @@ jobs:
     steps:
       - uses: DevCycleHQ/test-harness@main
         env:
-          SDK_CAPABILITIES: '["cloud","edgeDB","clientCustomData","multithreading","defaultReason","etagReporting","lastModifiedHeader","sdkConfigEvent","clientUUID"]'
+          SDK_CAPABILITIES: '["cloud","edgeDB","clientCustomData","multithreading","defaultReason","etagReporting","lastModifiedHeader","sdkConfigEvent","clientUUID","v2Config"]'
```
JamieSinn, line 15: this doesn't actually work properly - but good futureproofing

JamieSinn (approved):

.github/workflows/run-test-harness.yml
suthar26: i think it worked, tests started passing

PR #272, opened 2024-10-24
chore: update automation user

JamieSinn:


.github/workflows/test_examples.yml
```diff
@@ -9,15 +9,15 @@ jobs:
     name: Test Examples
     runs-on: ubuntu-latest
     env:
-      DEVCYCLE_SERVER_SDK_KEY: ${{ secrets.DEVCYCLE_SERVER_SDK_KEY }}
-      DEVCYCLE_VARIABLE_KEY: go-example-tests
+      DEVCYCLE_SERVER_SDK_KEY: dvc_server_token_test
```
jonathannorris, line 12: why fake keys now?

JamieSinn: they're not fake - but for the purposes of what we're testing (basic config parsing/eval) they suffice - it's our "public" test token we use throughout.

jonathannorris (approved):

PR #280, opened 2025-03-05
feat: Openfeature Provider Update

JamieSinn:


example/openfeature/main.go
```diff
@@ -24,10 +25,22 @@ func main() {
...
-	if err := openfeature.SetProvider(dvcClient.OpenFeatureProvider()); err != nil {
+	// Wait for the client to be initialized
+	for {
+		time.Sleep(time.Millisecond * 100)
+		event := <-inithandler
+		if event.EventType == api.ClientEventType_Initialized {
+			break
+		}
+	}
+	if err = openfeature.SetProvider(dvcClient.OpenFeatureProvider()); err != nil {
```
jonathannorris, line 43: There is no `SetProviderAndWait` method on Go yet?

JamieSinn: Nope - "await" as a concept doesn't exist in go

example/openfeature/main.go
```diff
@@ -24,10 +25,22 @@ func main() {
 		RequestTimeout:               time.Second * 10,
 		DisableAutomaticEventLogging: false,
 		DisableCustomEventLogging:    false,
+		ClientEventHandler:           inithandler,
+	}
+	dvcClient, err := devcycle.NewClient(sdkKey, &dvcOptions)
```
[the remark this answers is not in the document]
JamieSinn: Sure

openfeature_provider.go
```diff
@@ -313,7 +359,29 @@ func toOpenFeatureError(err error) openfeature.ResolutionError {
 	return openfeature.NewGeneralResolutionError(err.Error())
 }
 
-func createUserFromEvaluationContext(evalCtx openfeature.FlattenedContext) (User, error) {
+func createEventFromEventDetails(eventName string, details openfeature.TrackingEventDetails) Event {
+	event := Event{
+		Type_:       "customEvent",
+		CustomType:  eventName,
+		UserId:      "",
```
jonathannorris, line 366: shouldn't this come from the context?

JamieSinn: No - our user gets built separately and then combined later for the actual request - as they're paired - https://github.com/DevCycleHQ/go-server-sdk/blob/main/bucketing/event_queue.go#L446

example/openfeature/main.go
```diff
@@ -24,10 +25,22 @@ func main() {
...
+		log.Fatalf("Failed to create DevCycle client: %v", err)
 	}
-	dvcClient, _ := devcycle.NewClient(sdkKey, &dvcOptions)
 
-	if err := openfeature.SetProvider(dvcClient.OpenFeatureProvider()); err != nil {
+	// Wait for the client to be initialized
+	for {
+		time.Sleep(time.Millisecond * 100)
+		event := <-inithandler
+		if event.EventType == api.ClientEventType_Initialized {
```
jonathannorris, line 39: would it make more sense to use the `initialized` method on the provider?

jonathannorris: or actually this: https://openfeature.dev/docs/reference/technologies/server/go#eventing

JamieSinn: No - because we need a guaranteed return status and this event is the guarantee - I was fighting race conditions with the initialized method

jonathannorris: Does the OF one not guarantee to return?

JamieSinn: nope - it can return an error

jonathannorris: yea so can we catch the error?

jonathannorris: yea all I'm getting at here, is that we really should be trying to have these examples use the OpenFeature "native" way as much as possible, as that's what our customers will expect when they come to us for OpenFeature SDKs.

jonathannorris (approved):

example/openfeature/main.go
JamieSinn: Yep - agreed - though our initialization handling across all our SDKs is tricky and handled differently per language.

PR #288, opened 2025-05-30
feat: eval reasons

JamieSinn:
Add eval reasons to aggVariableEvaluated and aggVariableDefaulted

COR-3428

### DIFF
--- a/bucketing/event_queue.go
+++ b/bucketing/event_queue.go
@@ -83,9 +83,6 @@ func (agg *AggregateEventQueue) BuildBatchRecords(platformData *api.PlatformData
 					metaData := make(map[string]interface{})
 					evalMetadata := make(map[string]int64)
 					if _type == api.EventType_AggVariableDefaulted {
-						if _, ok := evalMetadata[string(api.EvaluationReasonDefault)]; !ok {
-							evalMetadata[string(api.EvaluationReasonDefault)] = 1
-						}
 						evalMetadata[string(api.EvaluationReasonDefault)]++
 						metaData["eval"] = evalMetadata
 						metaData["_variation"] = api.EvaluationReasonDefault
@@ -116,7 +113,9 @@ func (agg *AggregateEventQueue) BuildBatchRecords(platformData *api.PlatformData
 						metaData["eval"] = evalMetadata
 					}
 					event.MetaData = metaData
-					aggregateEvents = append(aggregateEvents, event)
+					if event.Value > 0 {
+						aggregateEvents = append(aggregateEvents, event)
+					}
 				}
 			}
 		}
--- a/bucketing/event_queue.go
+++ b/bucketing/event_queue.go
@@ -79,12 +79,10 @@ func (agg *AggregateEventQueue) BuildBatchRecords(platformData *api.PlatformData
 						FeatureVars: emptyFeatureVars,
 						ClientDate:  time.Now(),
 					}
-					var metaData map[string]interface{}
-					var evalMetadata map[EvaluationReason]int64
+					metaData := make(map[string]interface{})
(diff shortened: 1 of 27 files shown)


client.go
```diff
@@ -589,7 +589,7 @@ func (c *Client) performRequest(
 			return attempt <= 5, err
 		}
 		responseBody, err = io.ReadAll(httpResponse.Body)
-		httpResponse.Body.Close()
+		_ = httpResponse.Body.Close()
```
kaushalkapasi, line 592: since we are ignoring the response from the method, what's the reason for this change?

bucketing/event_queue.go
```diff
@@ -34,12 +35,14 @@ type userEventData struct {
 // map event type -> event target
 // map event target -> feature id
 // map feature id -> variation id
+// map variation -> eval reason count
 // For Evaluation Events:
-// ["aggVariableEvaluated"]["somevariablekey"]["feature_id"]["variation_id"] = 1
+// ["aggVariableEvaluated"]["somevariablekey"]["feature_id"]["variation_id"]["eval reason"] = 1
 // For Defaulted Events:
-// ["aggVariableDefaulted"]["somevariablekey"]["defaulted"][DEFAULT_REASON] = 1
+// ["aggVariableDefaulted"]["somevariablekey"]["DEFAULT"]["DEFAULT"]["DEFAULT_REASON"] = 1
```
kaushalkapasi, line 42: why is the `DEFAULT` doubled here?

JamieSinn: The normal aggEval map is mapped for `featureId` then `variationId` - in the default case we have neither - and to benefit from the aggregation efficiency - I just set the map key to be the same for each variation/feature id.

kaushalkapasi: makes sense, that way both entries have the same "depth" so we can iterate using the same logic

bucketing/event_queue.go
```diff
@@ -68,22 +71,24 @@ func (agg *AggregateEventQueue) BuildBatchRecords(platformData *api.PlatformData
...
+						ClientDate:  time.Now(),
 					}
-					var metaData map[string]interface{}
+					metaData := make(map[string]interface{})
+					evalMetadata := make(map[string]int64)
 					if _type == api.EventType_AggVariableDefaulted {
-						metaData = map[string]interface{}{
-							"defaultReason": variation,
-						}
+						metaData["evalDetails"] = variation
```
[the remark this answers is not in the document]
JamieSinn: The variation in this case is "DEFAULT"

client_native_bucketing.go
```diff
@@ -121,6 +127,10 @@ func (n *NativeLocalBucketing) Variable(user User, variableKey string, variableT
 			Value: resultValue,
 		},
 		IsDefaulted: false,
+		Eval: api.EvalDetails{
+			Reason:   evalReason,
+			TargetId: "",
```
kaushalkapasi, line 132: this should have a defined target id instead of `""` since this means the bucketing was successful?

JamieSinn: We have no target id because we have no targeting rule ids at all as part of the model on the datatype

client_test.go
```diff
@@ -193,10 +193,14 @@ func TestClient_VariableLocal(t *testing.T) {
 			Key:   "test",
 			Type_: "Boolean",
 			Value: true,
+			Eval: api.EvalDetails{
+				Reason: api.EvaluationReasonSplit,
```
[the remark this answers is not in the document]
JamieSinn: Details is empty here

kaushalkapasi (approved):

PR #289, opened 2025-06-05
fix: Handle hard limit for variation targeting

JamieSinn:


bucketing/model_target_test.go
```diff
@@ -82,7 +168,7 @@ func TestAudience_Parsing(t *testing.T) {
 		Id: "2d61e8001089444e9270bc316c294828",
 	}, audience)
 
-	filters := audience.NoIdAudience.Filters.Filters
+	filters := audience.Filters.Filters
```
kaushalkapasi, line 171: why this change?

JamieSinn: Coalesce - it's a passthrough field

PR #291, opened 2025-07-15
Fix possible deadlock on initialization and flushing+config set

JamieSinn:
It's plausible to cause a deadlock if the timing on fetching a config + initializing the event queue. This can happen on initialization as well as flushing events with high concurrency.

Also rename the event queue mutex for better clarity.


bucketing/config_manager.go
```diff
@@ -52,14 +52,17 @@ func GetRawConfig(sdkKey string) []byte {
...
 
 func SetConfig(rawJSON []byte, sdkKey, etag, rayId, lastModified string, eventQueue ...*EventQueue) error {
-	configMutex.Lock()
-	defer configMutex.Unlock()
 	config, err := newConfig(rawJSON, etag, rayId, lastModified)
 	if err != nil {
 		return err
 	}
+
+	configMutex.Lock()
```
kaushalkapasi, line 60: should the mutex lock start before the new config initialization (line 55)

.github/workflows/bench.yml
```diff
@@ -16,12 +16,12 @@ jobs:
       - name: Run benchmark
         run: CGO_ENABLED=0 go test -bench=^BenchmarkClient_VariableSerial -run=^# -benchDisableLogs | tee bench_native_output.txt
       - name: Download previous benchmark data
-        uses: actions/cache@v4
+        uses: actions/cache@v4.2.3
```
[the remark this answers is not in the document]
JamieSinn: Strict versions. It's a git tag

bucketing/config_manager.go
JamieSinn: No because it's just locking the map itself.

PR #295, opened 2025-07-24
feat: add-config-metadata 

suthar26:
adds config metadata to be used in hooks

### DIFF
--- a/api/model_sse.go
+++ b/api/model_sse.go
@@ -1,7 +1,9 @@
 package api
 
 type MinimalConfig struct {
-	SSE *SSEHost `json:"sse,omitempty"`
+	SSE         *SSEHost     `json:"sse,omitempty"`
+	Project     *Project     `json:"project,omitempty"`
+	Environment *Environment `json:"environment,omitempty"`
 }
 
 type SSEHost struct {
--- a/client.go
+++ b/client.go
@@ -308,6 +308,7 @@ func (c *Client) Variable(userdata User, key string, defaultValue interface{}) (
 			User:         userdata,
 			Key:          key,
 			DefaultValue: defaultValue,
+			Metadata:     c.GetMetadata(),
 		}
 		// Run before hooks and catch any errors
 		var hookError error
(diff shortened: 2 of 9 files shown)


config_metadata_test.go
```diff
@@ -0,0 +1,380 @@
...
+		resp := httpmock.NewBytesResponse(200, modifiedConfig)
+		resp.Header.Set("ETag", "test-etag-123")
+		resp.Header.Set("Last-Modified", "Wed, 21 Oct 2015 07:28:00 GMT")
+		resp.Header.Set("Cf-Ray", "test-ray-123")
+		resp.Header.Set("Content-Type", "application/json")
+		return resp, nil
+	}
+
+	const CONFIG_URL_FORMAT = "https://config-cdn.devcycle.com/config/v2/server/%s.json"
+	httpmock.RegisterResponder("GET", fmt.Sprintf(CONFIG_URL_FORMAT, sdkKey), responder)
```
JamieSinn, line 44: https://github.com/DevCycleHQ/go-server-sdk/blob/main/testing_helpers_test.go has helpers to make this cleaner

client.go
```diff
@@ -308,6 +308,7 @@ func (c *Client) Variable(userdata User, key string, defaultValue interface{}) (
 			User:         userdata,
 			Key:          key,
 			DefaultValue: defaultValue,
+			Metadata:     c.GetMetadata(),
```
JamieSinn, line 311: Why is this on the variable call instead of the client? This doesn't change per eval - and you're going to continuously allocate more memory in the variable response for no reason

api/model_sse.go
```diff
@@ -1,7 +1,9 @@
 package api
 
 type MinimalConfig struct {
-	SSE *SSEHost `json:"sse,omitempty"`
+	SSE         *SSEHost     `json:"sse,omitempty"`
+	Project     *Project     `json:"project,omitempty"`
+	Environment *Environment `json:"environment,omitempty"`
```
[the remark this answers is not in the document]
suthar26: are you suggesting to have the data here instead of the ConfigMetadata type?

client.go
```diff
@@ -590,12 +590,23 @@ func (c *Client) EventQueueMetrics() (int32, int32, int32) {
...
+		return ConfigMetadata{}, fmt.Errorf("config metadata not available - config not loaded")
 	}
-	return nil
+
+	return ConfigMetadata{
+		ConfigETag:         metadata.ConfigETag,
+		ConfigLastModified: metadata.ConfigLastModified,
+		Project:            metadata.ProjectMetadata,
+		Environment:        metadata.EnvironmentMetadata,
+	}, nil
```
JamieSinn, line 609: don't create a new variable - just access it

JamieSinn (approved):

PR #305, opened 2026-02-05
Stop preallocating event queue

mrdg:
Please see https://github.com/DevCycleHQ/go-server-sdk/issues/304 for context


JamieSinn:
We don't have a CLA in place - but #306 incorporates the goal here!

PR #310, opened 2026-08-05
chore(deps): bump open-feature/go-sdk to v1.17.2

jonathannorris:
## Summary
- Bumps `github.com/open-feature/go-sdk` from v1.14.1 to v1.17.2
- Removes `DevCycleProvider.Status()`, which is dead code as of go-sdk v1.16.0
- `go build`, `go vet`, and `go test ./...` all pass

### Notes
v1.16.0 (spec 0.8.0) made providers stateless and dropped `Status()` from `openfeature.FeatureProvider`, so the SDK no longer calls it. It was added in #280 only to satisfy the old interface, is not in the README or any example, and the only callers were its own tests. It also ignored `initialized()` and returned `ReadyState` before init completed. Consumers wanting provider state should use `openfeature.Client.State()`. `TestOFDevCycleProvider_Close` now asserts via `internalFullClient.closed()`, so coverage is unchanged.

Transitive: `testify` v1.10.0 to v1.11.1 (forced by MVS), and `go.uber.org/mock` is now a build-time indirect dep because go-sdk ships its mocks as non-test files.


suthar26 (approved):

openfeature_provider.go
```diff
@@ -60,14 +60,6 @@ func (p DevCycleProvider) Track(ctx context.Context, trackingEventName string, e
...
 }
 
-// Status expose the status of the provider
-func (p DevCycleProvider) Status() openfeature.State {
-	if p.Client.closed() {
-		return openfeature.FatalState
-	}
-	return openfeature.ReadyState
-}
-
```
JamieSinn, line 70: Removing this functionality requires a major bump. Just keeping it around and marking deprecated is safer.

openfeature_provider_test.go
```diff
@@ -19,10 +19,9 @@ func TestOFDevCycleProvider_Metadata(t *testing.T) {
 
 func TestOFDevCycleProvider_Close(t *testing.T) {
 	provider := getProviderForConfig(t, false)
-	require.Equal(t, openfeature.ReadyState, provider.Status())
+	require.False(t, provider.internalFullClient.closed())
 	provider.Shutdown()
-	state := provider.Status()
-	require.Equal(t, openfeature.FatalState, state)
+	require.True(t, provider.internalFullClient.closed())
```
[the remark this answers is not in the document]
JamieSinn: This is the more correct way to do this

openfeature_provider.go
jonathannorris: I had Claude look into this, and I don't think we need to make it a major version, claude's research: 

yea it's an exported method so I agree it's technically breaking. Worth noting the history though: `Status()` was part of the `StateHandler` interface (`Init`/`Shutdown`/`Status`) up to v1.13.0, and v1.14.0 dropped it. We added `Status()` in #280, which is the same commit that pinned v1.14.1, so it was dead on arrival: no released version of this SDK has shipped a go-sdk that calls it, and it was never documented for direct use.

happy to keep it with a `// Deprecated:` marker instead if you'd rather play it safe, it's only a few lines. I lean toward removing since it has never done anything, but your call.

JamieSinn: I would rather play it safe, just to avoid any potential version incompatibilities if someone didn't update their openfeature SDK to latest, etc. 

It's a small chance - but better safe than sorry - especially because we don't know what customers may be using

JamieSinn (approved):

openfeature_provider.go
jonathannorris: yea from what I can see this code was never executed because the version of the OF SDK required for using the provider never used this function, I really don't see a point in bumping a major version for that.

